Issue #108, opened 2014-06-06
Group names are not localized


monsta:
Originally reported at https://bugs.launchpad.net/bugs/1327107

![capture-1](https://cloud.githubusercontent.com/assets/5138986/3198340/15d67a96-ed58-11e3-8d66-e389dc05b3f6.png)

tarakbumba:
I think there is no need to wait for 1.12. There is already a proposed fix (pull request) :https://github.com/rezso/mate-control-center/commit/5ebc88406f51fe36bfb03319f1d4fb5ed8888b55

I have tested that and i can confirm that pull request fixes this issue.

monsta:
@stefano-k: what do you think about the mentioned fix?

stefano-k:
@monsta if you tried it and it works, it's ok for me

monsta:
@stefano-k: I confirm it works fine. Tested it in Debian Jessie with m-c-c 1.8.3. It applies cleanly to this version, so I think it makes sense to push it to 1.8 branch. It won't add any new dependencies because m-c-c already depends on mate-menus.

Moreover, it's even applicable to 1.6.x versions :) I have MATE 1.6 on my LMDE machine here. A simple patching of ```/etc/xdg/menus/matecc.menu``` did the trick.

PR #140, opened 2015-01-25
Use MATE's own .directory files in matecc

rezso:
First: Personal.directory, InternetAndNetwork.directory, Hardware.directory and System.directory shipped by gnome-menus, which is causes an unwanted gnome-menus dependency.
Second: because of different gettext domains, the translations are unused.

### DIFF
--- a/capplets/network/mate-network-properties.c
+++ b/capplets/network/mate-network-properties.c
@@ -41,14 +41,14 @@
  */
 
 /*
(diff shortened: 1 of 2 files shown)


tarakbumba:
This pull request fixes #108

flexiondotorg:
@stefano-k Please review this for inclusion in MATE 1.10

PR #152, opened 2015-03-25
libslab: do not reap child when launching desktop items

monsta:
should be used after https://github.com/mate-desktop/mate-desktop/pull/166 is applied


NiceandGently:
confirmed, that fixes the issue.

NiceandGently:
@stefano-k 
please review for 1.10

PR #157, opened 2015-05-08
UI GTK3: fix keyboard and mouse properties windows

NiceandGently:
- this fixes unusable keyboard properties window
- reduce size of mouse properties window


monsta:
@flexiondotorg: I've tested it - works fine in current Debian Testing. I think it can be merged.

flexiondotorg:
Thanks for your contribution.

PR #167, opened 2015-07-21
keyboard properties: don't set parent widget on show layout dialog

monsta:
fixes a GTK+ warning


infirit:
How about https://github.com/GNOME/gnome-control-center/commit/eac5028a7598b40941d565023aff5e372d7598aa

monsta:
Hmm... that leads to show layout dialog not appearing in window list on panel... kinda inconsistent with the same dialog launched via the layout icon in tray.

PR #176, opened 2015-08-06
use cairo to render sample font previews

dnk:
Gtk3 version does not show font previews.
this change fix it.

### DIFF
--- a/capplets/appearance/appearance-font.c
+++ b/capplets/appearance/appearance-font.c
@@ -32,6 +32,7 @@
 
 #include <glib/gi18n.h>
 #include <gio/gio.h>
+#include <cairo-ft.h>
 
 #include "capplet-util.h"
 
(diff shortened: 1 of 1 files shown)


infirit:
Thx, works here on Gtk2 MATE. The only thing I noticed is that the boxes are smaller even-though the a font size of 18 is the correct size, see https://imgur.com/Ya2F5tp,G2meTCd

There is just a small whitespace nitpick [here[(https://github.com/mate-desktop/mate-control-center/pull/176/files#diff-f5452505139a7cf306668ed08fa72a91R236).

Nice work :smile:

NiceandGently:
congrates, works well with gtk3,....another big step forward...
thank you

monsta:
Thanks a lot, works great in both GTK+ builds.

The only issue is the one that @infirit mentioned - font preview boxes are smaller than before.
Here's how it looks in GTK+3 build:

Before:

appearance-1
appearance-1a

After:

appearance-2 
appearance-2a

monsta:
I guess this one should be closed as https://github.com/mate-desktop/mate-control-center/pull/181 is merged.

PR #184, opened 2015-08-22
GTK3 Font-viewer: fix some deprecations and warnings

NiceandGently:
This reduce the warnings a lot.
Note, the ' deprecated gtk_widget_modify_bg' commit results in another deprecation (gtk_widget_override_background_color),
but better a deprecation warning from 3.16.0 as from 3.0.0.
Also, the horizontal size of the window can be reduced a bit, which comes from port to GtkGrid.

### DIFF
--- a/font-viewer/font-view.c
+++ b/font-viewer/font-view.c
@@ -1,6 +1,10 @@
-/* -*- mode: C; c-basic-offset: 4 -*-
- * fontilus - a collection of font utilities for MATE
+/* -*- mode: C; c-basic-offset: 4 -*- */
+
+/*
+ * font-view: a font viewer for MATE
+ *
  * Copyright (C) 2002-2003  James Henstridge <james@daa.com.au>
+ * Copyright (C) 2010 Cosimo Cecchi <cosimoc@gnome.org>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


monsta:
We target GLib >= 2.36 these days, so you can safely remove that 2.32 check :smile:

NiceandGently:
I think the port to sushi font reader needs to be done first.
For this only the gtk3 port GTKTable to GtkGrid is needed.
After that i can re-work the deprecation fixes.

font-viewer/font-view.c
```diff
@@ -246,22 +250,38 @@ add_row (GtkWidget *table,
...
+    gtk_widget_set_halign (name_w, GTK_ALIGN_END);
+    gtk_widget_set_valign (name_w, GTK_ALIGN_START);
+#else
     gtk_misc_set_alignment (GTK_MISC (name_w), 0.0, 0.0);
+#endif
     gtk_label_set_use_markup (GTK_LABEL (name_w), TRUE);
 
+#if GTK_CHECK_VERSION (3, 4, 0)
+	gtk_container_add (GTK_CONTAINER (grid), name_w);
+#else
```
infirit: small indentation problem

font-viewer/font-view.c
```diff
@@ -307,10 +354,18 @@ add_face_info (GtkWidget *table,
     GFileInfo *info;
     PS_FontInfoRec ps_info;
 
+#if GTK_CHECK_VERSION (3, 4, 0)
+	add_row (grid, _("Name:"), face->family_name, FALSE, FALSE);
```
infirit: indentation

PR #188, opened 2015-08-29
mouse: Use GSourceFunc instead of GtkFunction

pudh4418:
GtkFunction has been deprecated since version 2.24, and now we only
support GTK+ 2.0 version >= 2.24.


infirit:
merged thx

PR #192, opened 2015-08-30
Restore Interface tab in Appearance capplet

Exalm:
I found this thing pretty useful, so I decided to restore it.
As a bonus, help (mate-appearance-properties -h) still mentions the Interface tab, so this also fixes that. :)

Screenshot:
default

Note that I only tested it with GTK2. If anyone tell me where I can find a live cd with GTK3-based MATE, I can test it there as well.

### DIFF
--- a/capplets/appearance/Makefile.am
+++ b/capplets/appearance/Makefile.am
@@ -16,6 +16,8 @@ mate_appearance_properties_SOURCES = \
 	appearance-themes.h \
 	appearance-style.c \
 	appearance-style.h \
+	appearance-ui.c \
+	appearance-ui.h \
 	appearance-support.c \
 	appearance-support.h \
 	mate-wp-info.c \
--- a/capplets/appearance/appearance-main.c
+++ b/capplets/appearance/appearance-main.c
@@ -25,6 +25,7 @@
 #include "appearance-font.h"
 #include "appearance-themes.h"
 #include "appearance-style.h"
+#include "appearance-ui.h"
 #include "appearance-support.h"
 #include "theme-installer.h"
 #include "theme-thumbnail.h"
@@ -188,6 +189,7 @@ main (int argc, char **argv)
   desktop_init (data, (const gchar **) wallpaper_files);
   g_strfreev (wallpaper_files);
   font_init (data);
+  ui_init (data);
 
   /* init support for other window managers */
   support_init (data);
--- a/capplets/appearance/appearance-ui.c
+++ b/capplets/appearance/appearance-ui.c
@@ -33,11 +33,11 @@ show_handlebar (AppearanceData *data, gboolean show)
   g_object_ref (handlebox);
   g_object_ref (toolbar);
 
-  if (GTK_BIN (align)->child)
-    gtk_container_remove (GTK_CONTAINER (align), GTK_BIN (align)->child);
+  if (gtk_bin_get_child (GTK_BIN (align)))
+    gtk_container_remove (GTK_CONTAINER (align), gtk_bin_get_child (GTK_BIN (align)));
 
-  if (GTK_BIN (handlebox)->child)
-    gtk_container_remove (GTK_CONTAINER (handlebox), GTK_BIN (handlebox)->child);
+  if (gtk_bin_get_child (GTK_BIN (handlebox)))
+    gtk_container_remove (GTK_CONTAINER (handlebox), gtk_bin_get_child (GTK_BIN (handlebox)));
 
   if (show) {
     gtk_container_add (GTK_CONTAINER (align), handlebox);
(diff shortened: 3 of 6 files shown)


infirit:
One way to get the most obvious issues out of the was is to build with ```make CFLAGS+="-DGSEAL_ENABLE"```.

After below it builds under Gtk3 but.. I did a quick supervisual test and the toolbar text/icon and accelerator do not work. I think the changing of accelerator has been removed from Gtk3 not sure about the toolbar stuff.

```diff
diff --git a/capplets/appearance/appearance-ui.c b/capplets/appearance/appearance-ui.c
index d9f243a..da78892 100644
--- a/capplets/appearance/appearance-ui.c
+++ b/capplets/appearance/appearance-ui.c
@@ -33,11 +33,11 @@ show_handlebar (AppearanceData *data, gboolean show)
   g_object_ref (handlebox);
   g_object_ref (toolbar);
 
-  if (GTK_BIN (align)->child)
-    gtk_container_remove (GTK_CONTAINER (align), GTK_BIN (align)->child);
+  if (gtk_bin_get_child (GTK_BIN (align)))
+    gtk_container_remove (GTK_CONTAINER (align), gtk_bin_get_child (GTK_BIN (align)));
 
-  if (GTK_BIN (handlebox)->child)
-    gtk_container_remove (GTK_CONTAINER (handlebox), GTK_BIN (handlebox)->child);
+  if (gtk_bin_get_child (GTK_BIN (handlebox)))
+    gtk_container_remove (GTK_CONTAINER (handlebox), gtk_bin_get_child (GTK_BIN (handlebox)));
 
   if (show) {
     gtk_container_add (GTK_CONTAINER (align), handlebox);
```

Exalm:
Thanks, I've added your fix. But again, can you tell me where can I test GTK3 MATE? Building mate-control-center with --with-gtk=3.0 fails somewhere on mate-display-properties (both on Debian 8.1 with MATE 1.8 and on latest daily Ubuntu MATE with MATE 1.10 (the newest I could find)).

And I checked toolbar button style... It really does not work now. That's weird, since I clearly remember myself using both-horiz style with GTK3, and I doubt they will remove everything but the most "archaic" style ("both"). Maybe that's a bug in mate-settings-daemon, that it only applies that to GTK2.

infirit:
I know @monsta builds MATE gtk3 on debian so i am hoping he is willing to help with building :smile:.

NiceandGently:
There are pre-build mate gtk3 versions for fedora and archLinux.
https://copr.fedoraproject.org/coprs/raveit65/Mate-GTK3/
https://www.archlinux.org/packages/community/x86_64/mate-desktop-gtk3/

infirit:
btw, the keys not having any effect on Gtk3 imho is no a reason to reject this, the ui toggles the keys properly. But if you are willing to chase this down and fix it in mate-settings-daemon please do :smile:

infirit:
>GtkSettings:gtk-can-change-accels has been deprecated since version 3.10 and should not be used in newly-written code.

Maybe we should hide this if we build for >= 3.10..

Exalm:
Actually, menu icons (at least GtkImageMenuItem class) are now also deprecated, yet every single application (perhaps except GNOME apps?) uses them. (I hope MATE still uses them in GTK3 builds?) If MATE follows every GTK decision, it'll be pretty much like GNOME3 now, with CSD, symbolic icons everywhere, global menu etc.

monsta:
> Building mate-control-center with --with-gtk=3.0 fails somewhere on mate-display-properties (both on Debian 8.1 with MATE 1.8 and on latest daily Ubuntu MATE with MATE 1.10 (the newest I could find)).

You need to build the whole MATE with GTK+3... starting from the essential libraries like ```libmate-desktop``` and so on. I have a VM with Debian Testing and GTK+3 build of MATE, and I had to build it from scratch (using LXDE as a temporary desktop environment while I was doing that). I'll check this PR there.

Also @flexiondotorg has GTK+3 build for Arch Linux, so he might be able to test it as well.

> Actually, menu icons (at least GtkImageMenuItem class) are now also deprecated, yet every single application (perhaps except GNOME apps?) uses them. (I hope MATE still uses them in GTK3 builds?) If MATE follows every GTK decision, it'll be pretty much like GNOME3 now, with CSD, symbolic icons everywhere, global menu etc.

No, we don't follow every GTK+ decision :smile:
For example, buttons and menus with images are something we prefer to leave as is.

monsta:
I confirm that toolbar settings don't work in GTK+3 build.

Also... what "Editable menu shortcut keys" is supposed to do?

Exalm:
Yes, I've already tested that using Fedora 22 in a virtual machine and the Fedora repo @NiceandGently linked to. Those features have been removed from GTK3: https://github.com/mate-desktop/mate-settings-daemon/issues/64

As for the can-change-accels, it doesn't work in GTK3 as well, but on GTK2 it does this:
can-change-accels

Perhaps these switches (or the entire tab) should be hidden from GTK3 builds.

monsta:
> You just hover over a menu item and type the desired accel.

Ah so that's how it works - never seen that :smile:

infirit:
Squashed and merged.

PR #195, opened 2015-09-08
A much more organised categorisation of Control Center items.

flexiondotorg:
This organisation of categories has been well tested in Ubuntu MATE and makes use of all the available categories with no items being duplicated.

### DIFF
--- a/capplets/accessibility/at-properties/mate-at-properties.desktop.in.in
+++ b/capplets/accessibility/at-properties/mate-at-properties.desktop.in.in
@@ -6,6 +6,6 @@ Icon=preferences-desktop-accessibility
 Terminal=false
 Type=Application
 StartupNotify=true
-Categories=GTK;Settings;
+Categories=GTK;Settings;Accessibility;
 Keywords=mate-control-center;MATE;accessibility;features;
 OnlyShowIn=MATE;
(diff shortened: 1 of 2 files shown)


flexiondotorg:
These changes have also been tested on Linux Mint MATE 17.2.

flexiondotorg:
Any feedback for this? @infirit, @monsta @NiceandGently

monsta:
After the last commit I see all categories translated to Russian, so I don't mind :smile:

monsta:
Also we'd probably want @clefebvre's confirmation :smile:

NiceandGently:
I don't care very much if 'System' is renamed to 'Administration'. Maybe it's a better name for it.
+1

infirit:
Ok, so I don't see any of the problems like duplicate entries that this is fixing. But I do when the other commit, https://github.com/mate-desktop/mate-control-center/pull/197, is applied some things do end up in other. However I never had anything duplicate in the menu's before.

However, if this fixes issues for other ppl and it does not regress on my end, so I have no objection to this. But please update the commit message with what this is supposed to solve :smile:.

infirit:
I found a problem, the control-center application does not yet reflect these change.

monsta:
Rebased and merged, thanks :smile:

PR #197, opened 2015-09-08
RFC: Add categorisation of Preferences menu.

flexiondotorg:
This pull request is a proposal to add categories to the Preferences menu in MATE 1.12. This has been default in Ubuntu MATE for some time now and a screen shot is included to demonstrate what this looks like.

screenshot

The advantage of these categories is that the menu doesn't scroll off the screen on lower resolution displays and each item appears in a single category, making it easier to locate based on category.
This has been tested on Linux Mint MATE 17.2 without issue and I believe a variation of this preferences menu is available for Fedora also.

### DIFF
--- a/shell/Makefile.am
+++ b/shell/Makefile.am
@@ -27,9 +27,9 @@ directory_DATA = $(directory_in_files:.directory.in=.directory)
 @INTLTOOL_DIRECTORY_RULE@
 
 menudir = $(sysconfdir)/xdg/menus
-menu_DATA = matecc.menu
+menu_DATA = matecc.menu mate-preferences-categories.menu
 
-EXTRA_DIST = matecc.desktop.in.in matecc.directory.in matecc.menu
+EXTRA_DIST = matecc.desktop.in.in matecc.directory.in matecc.menu mate-preferences-categories.menu
 
 DISTCLEANFILES = matecc.desktop matecc.desktop.in matecc.directory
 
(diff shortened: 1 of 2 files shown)


infirit:
Even on my 21 inch screen the menu goes almost to the bottom. Even though it add an extra level it makes navigating easier, especially on smaller screens, :+1:.

NiceandGently:
Yes, i use this for a long time for fedora (Apr 2013), see https://github.com/NiceandGently/mate-preferences-menus/blob/master/mate-preferences-categories.menu
But i use this file in a subpackage which is in default mate-application yum/dnf group.
Means if someone don't like this old gnome menu style, he can uninstall the subpackage to have the normal menu back.
I remember that fedora gnome removed the categories from default with f10 or so, and started with a subpackage one release later after a lot of users complains about..

NiceandGently:
But why different sub entries for one category?
Ie, I would use X-MATE-PersonalSettings and X-GNOME-PersonalSettings and others in the same  sub entry, to make this file more simple and avoid duplications.
https://github.com/flexiondotorg/mate-control-center/commit/35ea8faefddbb856ef091944e36823ecbe76c868#diff-3dca00086ae828cbd7734f6ef4e0a84fR11

NiceandGently:
Btw. for testing you can simply copy this file to /etc/xdg/menus/

monsta:
Rebased and merged, thanks :smile:

PR #201, opened 2015-09-22
use offscreen windows for previews

dnk:
gtk 2.20+ provides offscreen window. use them for previews
minor fixes

### DIFF
--- a/capplets/common/theme-thumbnail.c
+++ b/capplets/common/theme-thumbnail.c
@@ -88,70 +88,6 @@ static int pipe_from_factory_fd[2];
 #define MARCO_THUMBNAIL_WIDTH  120
 #define MARCO_THUMBNAIL_HEIGHT  60
 
-/* This draw the thumbnail of gtk
- */
-#if GTK_CHECK_VERSION (3, 0, 0)
-static void draw_window_on_pixbuf(GtkWidget* widget, GdkPixbuf* pixbuf)
-#else
-static GdkPixmap* draw_window_on_pixbuf(GtkWidget* widget)
-#endif
-{
-	GdkVisual* visual;
-#if !GTK_CHECK_VERSION (3, 0, 0)
-	GdkPixmap* pixmap;
-#endif
-	GtkStyle* style;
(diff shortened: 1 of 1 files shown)


dnk:
there are still problems on gtk3 with theme previews :(

dnk:
now mate-appearance-properties correctly display previews event if thumbnails (~/.cache/thumbnails/normal/) are missing

raveit65:
It works with gtk3, but after cleaning the cache it creates only one thumbnail for one theme in cache directory, means i need to re-open mate-appearance-properties about 10 times for that all thumbnails are created. but maybe something is wrong with the resulting patch from this PR.
Anyways, great work :)

monsta:
I'd leave stock buttons as is even though they're considered deprecated since 3.10. I vote for using them until gnome guys really remove them from the code.
Besides, it will save us a new line in ```po/POTFILES.in```. Currently the build **breaks** because there are new translatable lines but this file isn't listed in ```po/POTFILES.in```:

```
The following files contain translations and are currently not in use. Please
consider adding these to the POTFILES.in file, located in the po/ directory.

capplets/common/theme-thumbnail.c
```

monsta:
Ok, and after removing these ```#if```s for stock buttons, I can confirm the weirdness @raveit65 reported:

> after cleaning the cache it creates only one thumbnail for one theme in cache directory, means i need to re-open mate-appearance-properties about 10 times for that all thumbnails are created.

And it's only in the main mate-appearance-properties window. As for the Customize Theme dialog, no thumbnails are created there at all. Controls, Colors, Window Border, Icons - all show only question marks, no matter how many times I restart the app. :disappointed:

monsta:
Hmm... tried this PR with GTK+2 build as well... decided to clean the cache here too - to see what will happen. Well it was even worse... not only Customize Theme dialog, but also the main mate-appearance-properties window has only question marks now. Restarting the app doesn't help.

Did I do something wrong? :confused:

monsta:
Closing as it really needs more work - for both GTK+ builds :confused:

PR #205, opened 2015-11-03
Fix dangling pointer errors

pudh4418:


raveit65:
Where do you see a dangling pointer?

pudh4418:
When the second time you double clicking the light bulb, it outputs some error messages like:
```
(mate-mouse-properties:2740): GLib-CRITICAL **: Source ID 190 was not found when attempting to remove it
```
It is because after removing the source, the source id variable still contains old value. So next time it will try to remove an already removed source. This is a dangling pointer error.
So just simply reset these variables to zero after removing the source.

raveit65:
yeah, confirmed

raveit65:
merged https://github.com/mate-desktop/mate-control-center/commit/fa57391891309545aebc94fae4d79f6fae683659
Thank you

PR #211, opened 2015-12-22
Use GDK convenient function

pudh4418:
Use gdk_x11_get_xatom_by_name (), and take advantage of GDK's caches.


monsta:
I've noticed ```XInternAtom``` is still used after that in ```mate-mouse-properties.c``` - should we change it there as well?

PR #214, opened 2016-01-05
make all three window focus modes selectable (fixes *marco* #140)

frimam:
This clears the issue #140 in marco and the third point on the roadmap for 1.14.  I hope it's correct that I added my name to the list at the top.

### DIFF
--- a/capplets/windows/mate-window-properties.c
+++ b/capplets/windows/mate-window-properties.c
@@ -155,10 +155,18 @@ mouse_focus_changed_callback (GSettings *settings,
                               const gchar *key,
                               gpointer user_data)
 {
-    gtk_toggle_button_set_active (GTK_TOGGLE_BUTTON (focus_mode_checkbutton),
-                                  g_settings_get_enum (settings, key) != FOCUS_MODE_CLICK);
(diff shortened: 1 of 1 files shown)


monsta:
Nice, I was going to work on this soon, but looks like this might save me some time :smile:
I'll check this later.

monsta:
Works good with a minor exception: when I try to set the focus mode via dconf-editor and the current mode is "click", I can't change it to "mouse". When I do that, it automatically becomes "sloppy" instead. After that, I can change "sloppy" to "mouse". I guess it's related to the way these nested checkboxes work here...

You can test it with dconf-editor as well (```org.mate.Marco.general -> focus-mode```).

@flexiondotorg @raveit65 do you wish to test this guys?

raveit65:
no, i trust you :)
I can test this with first 1.13 release

frimam:
I added a new commit to fix this behavior. The reason was that the settings are updated after each change of one of the check buttons so that there was the wrong setting saved. Now it works also next to external tools like dconf-editor.

monsta:
Cool, thanks, it works properly now :smile:

monsta:
Squashed into one commit and merged in https://github.com/mate-desktop/mate-control-center/commit/1b111a4e2698705026781b9bbc998bad019cf350.

PR #219, opened 2016-01-18
libslab: use GtkBox instead of GtkVBox in both GTK+ builds

monsta:
and fix deprecated usage of GtkHBox and GtkVBox in GTK+3 build

@flexiondotorg @sunweaver @gnugr: this commit changes the public interface of libslab in both GTK+ builds (due to transition of classes' parents and/or members from GtkVBox to GtkBox). So a rebuild of ```gnome-main-menu``` might be needed after this. However, I found out that g-m-m crashes with mate-panel 1.12 anyway, no matter if I rebuild it or not...

### DIFF
--- a/libslab/app-resizer.c
+++ b/libslab/app-resizer.c
@@ -275,13 +275,12 @@ app_resizer_size_allocate (GtkWidget * widget, GtkAllocation * allocation)
 }
 
 GtkWidget *
-app_resizer_new (GtkVBox * child, gint initial_num_columns, gboolean homogeneous,
+app_resizer_new (GtkBox * child, gint initial_num_columns, gboolean homogeneous,
 	AppShellData * app_data)
 {
 	AppResizer *widget;
 
 	g_assert (child != NULL);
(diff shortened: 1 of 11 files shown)


lukefromdc:
I just tested this, didn't find any problems with it at all. Mate-menus works fine, so do all the capplets in the control center I was able to test (not sure if I missed one while clicking through them though).

monsta:
Mate-menus will be alright; this change might affect only [gnome-main-menu](https://packages.qa.debian.org/gnome-main-menu) and only in GTK+2 build (since g-m-m still doesn't support GTK+3).

lukefromdc:
I don't think we can do anything about that package other than recommend
rebuilding. My theme support change to network-manager-gnome has been
totally ignored, not even a reply, leaving it a permanent fork in my usage.

monsta:
That's why I've notified the maintainers (it's maintained by Debian MATE team). However, as I also mentioned, this applet currently crashes with mate-panel 1.12 even without my changes. I'm not sure if it will even be maintained in the future.

PR #220, opened 2016-01-26
use offscreen window for previews

raveit65:
This is simply a rebase of previous PR from @dnk on top of 1.13.0 and for some reasons it works perfect with gtk3.
- no question marks in main window and editor
- all thumbnails are created in one rush after the cache was cleared.

I guess previous commits in master did the magic.

Of course it does not work with gtk2. Here i see everywhere question marks after the cache was cleared.
So we have 2 opportunities:
- fixes the commit for gtk2
- or simply use this for gtk3 only (pretty pefect easy, imo)

@monsta @dnk 
Please test and maybe you have a idea why this doen't work with gtk2

Note, i've reverted the GTKSTOCK_OPEN lines because this remove the open icon,
and for the deprecated gtk_widget_size_request fixes i used a single commit.

### DIFF
--- a/capplets/common/theme-thumbnail.c
+++ b/capplets/common/theme-thumbnail.c
@@ -94,16 +94,12 @@ static int pipe_from_factory_fd[2];
 
 /* This draw the thumbnail of gtk
  */
-#if GTK_CHECK_VERSION (3, 0, 0)
-static void draw_window_on_pixbuf(GtkWidget* widget, GdkPixbuf* pixbuf)
-#else
-static GdkPixmap* draw_window_on_pixbuf(GtkWidget* widget)
-#endif
-{
 #if !GTK_CHECK_VERSION (3, 0, 0)
+static GdkPixmap *
+draw_window_on_pixbuf(GtkWidget* widget)
+{
 	GdkVisual* visual;
 	GdkPixmap* pixmap;
-#endif
 	GtkStyle* style;
 	GdkScreen* screen = gdk_screen_get_default();
 	GdkWindow* window;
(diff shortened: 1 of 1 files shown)


monsta:
Cool! Works in GTK+3 for me, thanks :smile:
It also fixes some warning from cairo which always appeared on m-a-p start.

I'll check what can be done in GTK+2 later. Most probably I'll use the old code for GTK+2 (in order not to break things).

Some notes:
- I still get question mark on Adwaita and High Contrast (BTW, why do I have two High Contrast themes?). I think it's related to the warning that is shown below.

m-c-c-1

- in Customize -> Window Border tab there's wrong background color in some thumbnails. It's either white or black (color might change after a few theme switches). See Menta, Mint-X and some others. It should be gray for these themes.

m-c-c-2

monsta:
@lukefromdc: check this out :smile:

raveit65:
Questions marks with gnome themes is because gnome think they don't have to ship a metacity theme. M-a-p expected a gtk-theme, a icon-theme and a metacity-theme for creating the thumbnail. If one element isn't there than it failed of course.This you see also in gtk2 build. After  copying them from older gnome releases, the question marks are gone in one VM here.
- the background color of Window Border tab
I think it reads simply the "base" color from /usr/share/themes/BlueMenta/gtk-3.0/settings.ini,
which is #f7f7f7 (white).

raveit65:
You see 2 high Contrast theme because one i from mate-icon-theme and the other with question mark (no metacity theme) is from gnome.
See /usr/share/themes/ContrastHigh/indextheme, name is High Contrast but it's not translated.
I changed the name back to High Contrast a while ago because 'ContrastHigh' is not english ;)

raveit65:
commit is rebased on top of #221 and it works again.
Note, with gtk3 the first start of 'mate-appearance-properties -p theme' produced a lot of harmless warnings about 'unknow tag email, author, etc', after /home/rave/.cache/thumbnails/normal was cleared.
They will gone with a second start when the thumbnails are created.
Than only this warning is left:
Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.
I've no clue to fix that.

raveit65:
After looking at the cleaned commit, i noticed that GtkStyle will removed.
I'm pretty shure that using GtkStyle for gtk3 was the real issue.

lukefromdc:
I just tested this, works fine over here with gtk 3.19.7. Nice work on a nasty bug!

raveit65:
@monsta 
here is a stacktrace which shows what happened with gtk2.
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Bugs/backtrace-m-a-p-gtk2

monsta:
Ok, rebased and force-pushed changed commit, with GTK+2 code left intact. So it should work fine in both builds now.

raveit65:
Ok, updated commit works fine here but why not removing the obsolete ubuntu check for
gtk_window_set_has_resize_grip
https://github.com/mate-desktop/mate-control-center/commit/0fe0004f21832032602920d9c6f1074c1c3d9089#diff-4d46055f3215e6bc790af1905b3f1870R127
gtk_window_set_has_resize_grip is removed from gtk+ with 3.14. I can't imagine that a ubuntu version with 3.12/3.10 will ever get a mate-1.14 update + switching to gtk3 :)
Maybe we can add add the resize-grip function to libmate-desktop + using it in packages, to solve our issue with resizing windows.

monsta:
Check for Ubuntu's quirk is only in GTK+2 code now, so it does no harm in GTK+3 build. Actually I need to check if they still keep that patch... and check what it does, too (probably just turns resize grips always on).

raveit65:
OK, if it's for gtk2 why not using it in general for gtk2?
The resize-grip is very useful.

PR #244, opened 2016-07-24
 add some missing space  to the right of the bold username 

rogue-spectre:


raveit65:
Thank you.
Normal we accept only one PR on master and we cherry-pick them to current version if it is a bug fix, for me this is a bug fix.
For some reasons i got  a conflict with the other one, so i merged this one.

PR #247, opened 2016-08-01
Fix keybindings behavior when caps lock is on

pyhalov:
Ancient patch from Sun JDS, seems still to be relevant. Original author seems to be Arvind Samptur , if I read https://java.net/projects/solaris-desktop/lists/commits/archive/2013-03/message/0 correctly.

### DIFF
--- a/capplets/keybindings/eggcellrendererkeys.c
+++ b/capplets/keybindings/eggcellrendererkeys.c
@@ -426,7 +426,8 @@ static gboolean grab_key_callback(GtkWidget* widget, GdkEventKey* event, void* d
 		GDK_BUTTON2_MASK |
 		GDK_BUTTON3_MASK |
 		GDK_BUTTON4_MASK |
-		GDK_BUTTON5_MASK;
+		GDK_BUTTON5_MASK |
+		GDK_LOCK_MASK;
 
 	/* filter consumed/ignored modifiers */
 	if (keys->accel_mode == EGG_CELL_RENDERER_KEYS_MODE_GTK)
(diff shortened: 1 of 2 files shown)


monsta:
Would be nice to see the description of the problem first.

pyhalov:
Original bug report: "With Caps Lock off when you try to assign the same key to two or more different commands an error box will pop up preventing you from doing so. However if you turn caps lock on you will be able to assign the key to any number of different commands and some weird keycodes will appear as shown in the attached screenshots. If you delete all these keyboard shortcuts you will not be able to use the key which you assigned. But when you log out and in again no key will work. Also the backspace key will now be assigned to one of the shortcut commands.". Fortunately, I can't reproduce this behavior with or without this patch, as now you can't assign alphabetic shortcuts (like 'p').

monsta:
@flexiondotorg @raveit65 @lukefromdc @XRevan86 @sc0w
Guys, can you reproduce the mentioned issue?

lukefromdc:
This did NOT happen on my system, had to create a custom shortcut to test, then assign a combination already in use to it. Error dialog came up with or without caps lock set

monsta:
I suspect it only happened in some old version of Gnome 2, where, as mentioned above, it was allowed to assign a single alphabet key (without Ctrl, Alt or other modifiers) to a command. I don't even remember now if I ever used such an old version.

monsta:
The date of patch listed in the link above only confirms that: 2004-11-09.

monsta:
Ok, I see it had been fixed in another way, and we already have this code:
https://git.gnome.org/browse/gnome-control-center/commit/?id=09076c4e2765ca96aa89b47de1cb4cf3e1a65c06
https://git.gnome.org/browse/gnome-control-center/commit/?id=04d6b6038d3047bb48bdec1d4b1e11a29404d504

PR #248, opened 2016-08-01
pkexec doesn't exist on illumos/Solaris

pyhalov:


pyhalov:
This is a wrong solution, we just have to ship polkit. But mate-control-center could check for its existance...

PR #253, opened 2016-10-01
GTK3: port libunique->GtkApplication

lukefromdc:
All GTK3 builds for now, this can go 3-way to permit gtk3/libunique builds if there is any reason for a distro to have issues with using GtkApplication, or for a point/buildfix release. No new warnings at build or runtime that I could find.


sc0w:
what means `StartupNotify=false` on  `shell/matecc.desktop.in.in` ?

---------------------------------------------------------

If I close, I recieve this warning: (debian testing gtk 3.22.0)

`(mate-control-center:26802): Gtk-CRITICAL **: gtk_main_quit: assertion 'main_loops != NULL' failed`

-----------------------------------------------------------

I think you must to do something like this to work with gtk2

```patch
diff --git a/configure.ac b/configure.ac
index 8dcb34f..4b2df49 100644
--- a/configure.ac
+++ b/configure.ac
@@ -88,7 +88,6 @@ case "$with_gtk" in
        GTK_REQUIRED=3.0.0
        MATE_DESKTOP_REQUIRED=1.15.1
        LIBCANBERRA_API_VERSION=3
-       LIBUNIQUE_API_VERSION=3.0
        ENGINES_FOLDER="theming-engines"
        APPINDICATOR_API_VERSION=3
        ;;
@@ -160,7 +159,13 @@ COMMON_MODULES="gtk+-$GTK_API_VERSION >= $GTK_REQUIRED dnl
  mate-desktop-2.0 >= $MATE_DESKTOP_REQUIRED"
 PKG_CHECK_MODULES(CAPPLET, $COMMON_MODULES)
 PKG_CHECK_MODULES(MATECC, $COMMON_MODULES libmate-menu >= 1.1.0)
-PKG_CHECK_MODULES(MATECC_SHELL, $COMMON_MODULES libmate-menu unique-$LIBUNIQUE_API_VERSION)
+
+if test "$GTK_API_VERSION" = "3.0"; then
+  PKG_CHECK_MODULES(MATECC_SHELL, $COMMON_MODULES libmate-menu)
+else
+  PKG_CHECK_MODULES(MATECC_SHELL, $COMMON_MODULES libmate-menu unique-$LIBUNIQUE_API_VERSION)
+fi
+
 PKG_CHECK_MODULES(DBUS, dbus-1 dbus-glib-1)
 PKG_CHECK_MODULES(MATE_DESKTOP, mate-desktop-2.0 >= MATE_DESKTOP_REQUIRED)
 PKG_CHECK_MODULES(DEFAULT_APPLICATIONS_CAPPLET, libxml-2.0)

lukefromdc:
I'm not getting this warning but did get it prior to removing one instance of gtk_main_quit (), I guess more of those will have to go behind GTK2 only selectors (possibly all of them for the shell?).

Thanks for the configure.ac tips, I will test them out tomorrow. I was able in initial testing to complete a configuration for GTK2 but did not test a build as I am on GTK3. Often though a package built that way won't run (mix of GTK verrsion) but will still complete the build so I will run that test.

The startupnotify=false is necessary to avoid a stuck busy spinning cursor on a second invocation of m-c-c from the panel menus, just as it was needed in Caja for opening the trash or from the places menu.

lukefromdc:
https://github.com/sc0w,  can you retest to see if I found your warning? I checked the non-default gsettings options for closing the control center shell on opening a task, and got both the same warning and a failure to close the shell. Replaced gtk_main_quit with g_application_quit in 4 places for GTK3 only, fixing both those warnings and the shell closing behavior on opening tasks from within it. Both true and false gsettings options now work as before.

BTW, I found another issue with this option also present in master and not caused by or modified by my work: If you set the shell to close on opening a task and the task is one of the capplets, the terminal does not release. It does release if the task is a 3ed party app like emerald-theme-manager. Both the control center and the task disappear from the process list in mate-system-settings but the terminal does not release. Since it can release (as in 3ed party tasks), the capplets themselves are doing something wrong. None are changed by these commits, they are a matter for later and in any event this has no effect when the default options to keep the shell open are used.

sc0w:
your latest changes don't solve the warning for me:

`(mate-control-center:14973): Gtk-CRITICAL **: gtk_main_quit: assertion 'main_loops != NULL' failed`

> BTW, I found another issue with this option also present in master and not caused by or modified by my work: If you set the shell to close on opening a task and the task is one of the capplets, the terminal does not release. It does release if the task is a 3ed party app like emerald-theme-manager. Both the control center and the task disappear from the process list in mate-system-settings but the terminal does not release. Since it can release (as in 3ed party tasks), the capplets themselves are doing something wrong. None are changed by these commits, they are a matter for later and in any event this has no effect when the default options to keep the shell open are used.

just, you can report the new issue :)

lukefromdc:
How exactly are you closing mate-control-center? I cannot duplicate the
warning you are getting.

lukefromdc:
The only way I know how to close mate-control-center with the default gsettings value is to close the window. I normally keep recently-used.xbel disabled, so I get this warning and only this warning if run mate-control-center from terminal and immediately close it.  Capplets will create their own warnings on startup, things like "Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged" but never the gtk_main_quit warning.

sc0w:
> How exactly are you closing mate-control-center? I cannot duplicate the warning you are getting.

click on X or ALT+F4

lukefromdc:
I just tried both, clean close with no warnings. My build is over GTK 3.22.1,
glib2.50.0, and a rather heavily modded Debian Unstable but no local patches
applied to GTK, glib, or anything else I would expect to have a bearing on this.

sc0w:
@monsta @raveit65 please, can you test if you receive the warning when closing?

sc0w:
see https://github.com/mate-desktop/mate-power-manager/pull/199#issuecomment-236168171

same warning was solved

lukefromdc:
A build warning found one missed instance of gtk_main_quit for me, maybe you were hitting that. I can't use  gtk_widget_destroy on app_data because it does not come up as a gtk widget. If you still get the warning, I can try a quit callback with code to find and destroy some key widget but that's more complex and would require including one file (or a header containing it) in the other's build to access it

lukefromdc:
Title of last commit change to GTK3: fix missed instance of gtk_main_quit()

monsta:
Works fine here with both GTK+, thanks. Squashed and merged in https://github.com/mate-desktop/mate-control-center/commit/80d37475904c9e58b8a291bcd6d49867dbf23506, with a bit of indent corrections.

sc0w:
this is odd

`(mate-control-center:8770): Gtk-CRITICAL **: gtk_main_quit: assertion 'main_loops != NULL' failed`

@monsta 

@raveit65 

you don't experience this warning?

monsta:
Wait, it was fixed by @lukefromdc in his last commit... and I think I've squashed all the commits.

sc0w:
@monsta I have tested with the latest master branch with https://github.com/mate-desktop/mate-control-center/commit/80d37475904c9e58b8a291bcd6d49867dbf23506

you don't see the warning?

monsta:
No... I've tried various ways of closing the app, nothing...

sc0w:
ok, perfect :)

monsta:
Did the warning disappear by itself? :open_mouth:

sc0w:
I see the warning

I don't know what happen on my system xD

PR #262, opened 2016-12-12
fix uk translation

sggua:
It should be changed in all cases in all language files:
1. "Иньш" -> "Інш", "иньш" -> "інш";
2. "Кляв" -> "Клав", "кляв" -> "клав".
// Sergiy Gorichenko


monsta:
We can merge it, but it will be overwritten next time we sync from Transifex.
It's better to make changes on Transifex. For example, you can browse Ukrainian translations at https://www.transifex.com/mate/MATE/language/uk/

Issue #143, opened 2015-02-09
Unable to see font characters while using mate-font-viewer


shirishag75:
Dear Developer, 
Thank you for making mate desktop and mate-control-center.  

The issue at hand is I was trying to see a font file but am getting empty rectangular boxes while the same font is seen in all it's glory while viewing the same via gnome-font-viewer. 

Sharing how it looks under mate-font-viewer and gnome-font-viewer. 

![mate-font-viewer](https://cloud.githubusercontent.com/assets/727626/6101561/1c64d440-b04c-11e4-88f4-1a030563aa3a.png)

![gnome-font-viewer](https://cloud.githubusercontent.com/assets/727626/6101563/25104f5c-b04c-11e4-9972-68ac529d554e.png)

Look forward to a solution.
##

monsta:
I confirm that - but only with this particular font. Other fonts like Droid or Noto are displayed properly.

NiceandGently:
After i read those commit https://git.gnome.org/browse/gnome-font-viewer/commit/?id=d11b6ae9f8e1a62fe366e349793a38e7040aefb5
i think this lines in code needs to be adjusted.
https://github.com/mate-desktop/mate-control-center/blob/master/font-viewer/font-thumbnailer.c#L319

monsta:
It can't be applied... it's another part of code from another file (which we apparently don't have here).

infirit:
The function ``check_font_contains_text`` is in ``font-view.c`` and the only change compared to MATE is the calling FT_Select_Charmap. Below is an untested patch for this commit.

Where can I download this font? ``android emoji`` does not seam to be something I can install here.. 

```diff
diff --git a/font-viewer/font-view.c b/font-viewer/font-view.c
index cf3c651..eb1c1e3 100644
--- a/font-viewer/font-view.c
+++ b/font-viewer/font-view.c
@@ -61,6 +61,8 @@ draw_string (cairo_t *cr,
 static gboolean
 check_font_contain_text (FT_Face face, const gchar *text)
 {
+    FT_Select_Charmap (face, FT_ENCODING_UNICODE);
+
     while (text && *text) {
 	    gunichar wc = g_utf8_get_char (text);
```

NiceandGently:
@infirit 
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Bugs/android-emoji.ttf

infirit:
patch does not work, ignore it.

[...]

PR #265, opened 2017-01-08
re-work of font viewer

raveit65:
I finished all the work a while ago at 30 Aug 2015 here at https://github.com/raveit65/mate-control-center/commits/dev-font-viewer-gtk3
So please don't ask me about details why i used a commit, i have forgotten it.  ;-)
I started with a port to sushi font loader to fix https://github.com/mate-desktop/mate-control-center/issues/143
Than i back ported more and more commits from upstream.
I used them since the beginning for the gtk3 testing repo for fedora and with f25 for official builds.
And i got no negative feedback.
@lukefromdc 
Please test it too


lukefromdc:
I just tested it, seems to work fine and give good detail of font smoothing differences

monsta:
Wow, that's really huge. Basically, it's a new app. At least it would be so for most users who probably never even knew about mate-font-viewer (due to ```NoDisplay=true```). I wonder if we should really go this way and copy all the new functionality from gnome-font-viewer...

raveit65:
Why?
Having a overview window with all installed fonts is a improvement for me.

flexiondotorg:
I'd like to see the improved font viewer land and also think it should be made visible in the menus.

raveit65:
I squashed the sync with Sushi commits in one like suggested in irc.
Squashing the last sync commit isn't possible without editing a bunch of files.

raveit65:
Now i squashed the gd-main-toolbar commits.

raveit65:
Ok, i wouldn't squash more commits.

Issue #256, opened 2016-10-29
mate-appearance-properties: mounse pointer size slider greyed out


quite:
As per https://ubuntu-mate.community/t/support-for-cursors-with-multiple-sizes/1356/2 

This seems to be the case still. The pointer size slider (smaller/larger) is not enabled/disabled (grey out) to reflect whether the currently selected pointer theme supports different sizes or not. 
##

monsta:
Ok, it's fixed now, and I will also backport it to 1.16.

PR #266, opened 2017-01-12
Fix live updating of some theme interface elements

cwendling:
Fix list-specific live updates, like whether the theme can be deleted, or the available cursor sizes for the current theme.

This got lost in GSettings migration in b3e27b1d6a0f11c5835f5829bac3861e2147b17e.  The initial value update was restored in 070e7cb765a788c8e27ff3615d3ed36981d3518d.

Note: as signals handlers are called in the order they are registered, the specific one will run after the generic one, so the generic updates will have been applied.

----
The leaks fix is unrelated, and can be split out it you want.

### DIFF
--- a/capplets/appearance/appearance-style.c
+++ b/capplets/appearance/appearance-style.c
@@ -980,6 +980,7 @@ prepare_list (AppearanceData *data, GtkWidget *list, ThemeType type, GCallback c
   gchar *signal_name = g_strdup_printf("changed::%s", key);
   g_signal_connect (settings, signal_name,
       G_CALLBACK (treeview_gsettings_changed_callback), list);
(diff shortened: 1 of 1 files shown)


monsta:
Looks good to me. Cursor size slider is now properly updated, so https://github.com/mate-desktop/mate-control-center/issues/256 is fixed.

raveit65:
I see no reason not to merge it after you approved it.

monsta:
Ok, rebased and merged in https://github.com/mate-desktop/mate-control-center/commit/1a2b8e9f6b96cb2d6b768e901a126ff770b4b8bf and https://github.com/mate-desktop/mate-control-center/commit/28d95d764e6b5c95b1a0e5556cc183e28dd671ae.

PR #271, opened 2017-03-02
UI: avoid deprecated widgets

raveit65:
i hope all looks like before ;-)

### DIFF
--- a/capplets/about-me/mate-about-me-dialog.ui
+++ b/capplets/about-me/mate-about-me-dialog.ui
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.20.0 -->
 <interface>
-  <!-- interface-requires gtk+ 3.0 -->
+  <requires lib="gtk+" version="3.14"/>
+  <!-- interface-license-type gplv2 -->
+  <!-- interface-name mate-about-me-dialog -->
+  <!-- interface-description about me dialog window -->
+  <!-- interface-copyright MATE Developers -->
+  <!-- interface-authors Wolfgang Ulbrich -->
   <object class="GtkDialog" id="about-me-dialog">
     <property name="can_focus">False</property>
     <property name="border_width">5</property>
@@ -41,18 +47,19 @@
           </packing>
         </child>
         <child>
-          <object class="GtkVBox" id="vbox55">
+          <object class="GtkBox" id="vbox55">
             <property name="visible">True</property>
             <property name="can_focus">False</property>
             <property name="border_width">5</property>
+            <property name="orientation">vertical</property>
             <property name="spacing">12</property>
             <child>
-              <object class="GtkHBox" id="hbox58">
+              <object class="GtkBox" id="hbox58">
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="spacing">48</property>
                 <child>
-                  <object class="GtkHBox" id="hbox59">
+                  <object class="GtkBox" id="hbox59">
                     <property name="visible">True</property>
                     <property name="can_focus">False</property>
                     <property name="spacing">12</property>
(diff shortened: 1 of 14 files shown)


monsta:
```
$ mate-appearance-properties 

(mate-appearance-properties:19289): GLib-GObject-WARNING **: invalid (NULL) pointer instance

(mate-appearance-properties:19289): GLib-GObject-CRITICAL **: g_signal_connect_data: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed

(mate-appearance-properties:19289): GLib-GObject-CRITICAL **: g_object_ref: assertion 'G_IS_OBJECT (object)' failed

(mate-appearance-properties:19289): Gtk-CRITICAL **: gtk_bin_get_child: assertion 'GTK_IS_BIN (bin)' failed
```

monsta:
Mouse properties dialog has some labels centered - looks weird:

mate-mouse-oddities

raveit65:
I didn't saw that as german words are longer ;-)
I will align them to  the left side

monsta:
Use ```LC_ALL=C mate-mouse-properties``` for the default locale :slightly_smiling_face:

raveit65:
This is not the problem.
All those labels use
```
 <property name="halign">start</property>
```
ie for timeout_label
https://github.com/mate-desktop/mate-control-center/blob/master/capplets/mouse/mate-mouse-properties.ui#L669
If i use xalign=0 than they are at the left side.
But than glade complains about a deprecated widget.
Maybe glade is wrong here?
I recall that we use xalign for labels for replacing gtk_misc_alignment.

monsta:
Indeed, we use it in the code, and it's only for GTK+ >= 3.16. The docs say [the same](https://developer.gnome.org/gtk3/stable/GtkLabel.html#GtkLabel--xalign). It's not deprecated for sure...

raveit65:
Ok, i will verify if it works for 3.14 (RHEL) too, otherwise i have a upgrade problem for 1.18 if i use {x/y}align.

raveit65:
@monsta 
https://github.com/mate-desktop/mate-control-center/commit/84046ddf426fac7ac36b56fc488b830a8962f6a8
This works out of box with 3.14 and generate no runtime warnings.

monsta:
Good, but labels near sliders are still centered :slightly_smiling_face:

raveit65:
Which labels?
Please be specific.....

raveit65:
Those warnings are coming from removing deprecated 
```
<object class="GtkHandleBox" id="toolbar_handlebox">
```
Looks like the is an remnant in code itself.

raveit65:
Ok, i need to remove the deprecated GtkHandleBox here too.
https://github.com/mate-desktop/mate-control-center/blob/master/capplets/appearance/appearance-ui.c#L29

raveit65:
@monsta 
Can you please test this UI file before i commit.
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Test/appearance.ui
I didn't removed GtkHandleBox here, this needs to be done in code first.
Maybe replacing it with a GtkBox? But this seems to be a bit overdosed.
align --> GtkBox --> Toolbar
better
align --> Toolbar

raveit65:
@monsta 
please test https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Test/mate-mouse-properties.ui
...i don't think it looks better but i don't care ;-)

monsta:
Mouse UI seems ok.

monsta:
Warnings from appearance are gone. I didn't even know we had a handlebox there.

monsta:
Ah, I see. It was added along with the last tab in https://github.com/mate-desktop/mate-control-center/commit/57026bf522ee80d8fc3710ca33da73ff103f98c3. The code should react to ```toolbar-detachable``` key from ```org.mate.interface``` schema and make the sample toolbar detachable from the window... only that it doesn't happen, at least with GTK+ 3.18.9 here.

PR #274, opened 2017-03-24
keybindings: Allow to select an installed application through a dialog

cwendling:

PR #277, opened 2017-04-06
keybindings: Fix and add mnemonics in the custom command dialog

cwendling:
Also remove an unnecessary intermediate box, we can pack in the grid
directly.

Fix for breakage introduced in #274.

### DIFF
--- a/capplets/keybindings/mate-keybinding-properties.ui
+++ b/capplets/keybindings/mate-keybinding-properties.ui
@@ -97,6 +97,7 @@
                   <packing>
                     <property name="left_attach">1</property>
                     <property name="top_attach">0</property>
+                    <property name="width">2</property>
                   </packing>
                 </child>
                 <child>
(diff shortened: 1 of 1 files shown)


cwendling:
For best diff, check it together with the previous commit: `git diff 5d928407022afa0fd9e1c6798f4c93ba54418b4e^`

monsta:
These are keyboard shortcuts that you can see if you hold Alt - the corresponding letters in labels, buttons and other text widgets will be underlined. Pressing Alt + that letter activates it and makes widget do something. For example, Alt+B in this dialog will push "Browse applications..." button, and you will see 'B' underlined in button text.

monsta:
As for the situation you described - can you get out of it with Esc? It should cancel any editing action...

lukefromdc:
OK, the shortcuts show up if you hold alt down, and yes, esc will get you out of the situation where it is waiting for input and nothing responds to the mouse

PR #280, opened 2017-05-06
deprecations and warnings

raveit65:
Please test.


raveit65:
It seems the watermark is broken with gtk2 too, only the height of the GtkEntry is smaller in result it is less conspicuous.
m-c-c-search
The watermark will create in search-entry.c, with cairo and rsvg, but i have no idea to fix it there.

raveit65:
The watermark image itself is offset to the top, see https://github.com/mate-desktop/mate-control-center/blob/master/libslab/search-entry-watermark.svg
What we can do is to remove the usage of the image and use the find icon from icon-themes, what you see in caja searchbar for example.
What do you think?

PR #281, opened 2017-05-16
fix some deprecations

raveit65:
Please test.


lukefromdc:
Here's the difference in the sidebars, I noticed no other changes.

mcc_new_sidebar_deprecations2
mcc_old_sidebar

raveit65:
Thanks for testing.
Last commit is updated and should fix the fallout.
Please test again.

lukefromdc:
The fix works, the "change theme" and "set preferred applications" buttons now look like they always did

Issue #283, opened 2017-05-18
[GTK+3] m-c-c shell seems to miss icons from /usr/share/pixmaps


monsta:
It was first reported at https://github.com/mate-desktop/mate-icon-theme/issues/24, but it's not a mate-icon-theme issue. The problem is that if an icon theme doesn't have the needed icon, and `/usr/share/icons/hicolor/` doesn't have it too, the default icon isn't loaded from `/usr/share/pixmaps`.

The issue only happens on m-c-c start. It won't happen when m-c-c is running and you change the icon theme. See more details in https://github.com/mate-desktop/mate-icon-theme/issues/24.

Reproduced with:
- synaptic (`/usr/share/pixmaps/synaptic.png`)
- qt4-qtconfig (`/usr/share/pixmaps/qtconfig-qt4.png`)
- nvidia-settings (`/usr/share/pixmaps/nvidia-settings.png`)

PR #284, opened 2017-05-18
libslab: correct icon loading code

monsta:
fixes loading app icons from /usr/share/pixmaps when the icon theme
doesn't have an icon with chosen name

fixes https://github.com/mate-desktop/mate-control-center/issues/283


monsta:
I'll hold it for a while because on some systems it makes m-c-c shell ignore icon theme changes (have to restart it to load the new theme).

monsta:
Ok, this new commit should do it. Everyone please check that the icons in m-c-c shell properly change when the icon theme changes.

lukefromdc:
Works fine here, and icons change with the theme in the control center. Synaptic's icon is supplied by Synaptic itself, so it stays the same.

PR #289, opened 2017-07-02
mouse: warn about enabling multi-finger emulation

monnerat:
libinput is the new default driver for touchpads and unlike the synaptics driver, it disables software buttons when clickfinger is enabled.

This commit introduces a warning message about this situation in the touchpad dialog. This message only appears when some multi-finger click emulation is enabled.

See https://bugzilla.redhat.com/show_bug.cgi?id=1465115

Thanks for considering it.

### DIFF
--- a/capplets/mouse/mate-mouse-properties.c
+++ b/capplets/mouse/mate-mouse-properties.c
@@ -219,11 +219,30 @@ synaptics_check_capabilities (GtkBuilder *dialog)
 }
 
 static void
-comboxbox_changed_callback (GtkWidget *combobox, void *data)
+comboxbox_changed (GtkWidget *combobox, GtkBuilder *dialog, const char *key)
 {
 	gint value = gtk_combo_box_get_active (GTK_COMBO_BOX (combobox));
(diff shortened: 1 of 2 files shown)


lukefromdc:
Is there any way to invoke the dialog on a desktop? I do not have a modern laptop with a fancy touchpad, have only desktops (all development on them) and a very old netbook. I do not know how to use any of the touch features mentioned so have never explored much less enabled them, don't even know if they exist on old netbooks or not.

raveit65:
@monnerat 
I have added your patch to fedora builds for f25/26.
https://bodhi.fedoraproject.org/updates/FEDORA-2017-0a2df5b815
One last question, what happens if synaptic driver is in used?
Do we see the warning in GUI in this case?

monnerat:
Yes we do, This is why the warning says "may disable" instead of "disables".

raveit65:
Ok, i am fine with this.

raveit65:
Thank you

monnerat:
Thanks for merging :-)

PR #290, opened 2017-07-15
fix a few deprecations

raveit65:


raveit65:
Changes are for the locking keyboard screen itself. I used 1min for both timer settings in GUI for testing.

lukefromdc:
Locking function also works fine

PR #293, opened 2017-07-24
capplet-util: don't use deprecated gtk_show_uri

raveit65:
Please test.


raveit65:
Thanks for the review.

PR #295, opened 2017-08-02
avoid deprecated gdk_screen_make_display_name

sc0w:


raveit65:
It build just fine, but we can't really test the function of the file as it is only use by gnome-main-menu.
Gnome-main-menu is really EOL and not ported to gtk+-3.
I just prepared to remove document-tile.c and directory-tile.c in a local commit.

PR #298, opened 2017-08-07
don't use excessive string duplication

sc0w:
@monsta you did the same in https://github.com/mate-desktop/caja-extensions/commit/5fa187fd69c9a352bae38cbced4120dc942fc7a6

please test


sc0w:
I see no warnings, but after https://github.com/mate-desktop/mate-panel/pull/640 I decide to close this PR too

PR #317, opened 2018-01-08
Fix Window Border theme cannot be deleted

### LINKED ISSUE #316: Window Border theme cannot be deleted
#### Expected behaviour
The theme can be deleted.

#### Actual behaviour
The theme can not be deleted.

1. Open m-c-c → "Appearance" → Theme → Customize... → Window Border
2. Then, choose a theme to delete, it will be failed with tips: "Theme can not be deleted"

#### MATE general version
1.19

#### Package version
1.19.2-1

#### Linux Distribution
Fedora 27

zhangxianwei8:
The directory "marco-1" does not exist, replacing it  with
"metacity-1".

Fixes #316

Signed-off-by: Zhang Xianwei <zhang.xianwei8@zte.com.cn>

### DIFF
--- a/capplets/appearance/theme-installer.c
+++ b/capplets/appearance/theme-installer.c
@@ -310,10 +310,6 @@ mate_theme_install_real (GtkWindow *parent,
 					   theme_name, NULL);
 		break;
 	case THEME_MATE:
-		target_dir = g_build_path (G_DIR_SEPARATOR_S,
-					   g_get_home_dir (), ".themes",
-			 		   theme_name, NULL);
-		break;
 	case THEME_MARCO:
 	case THEME_GTK:
 		target_dir = g_build_path (G_DIR_SEPARATOR_S,
--- a/capplets/appearance/theme-util.c
+++ b/capplets/appearance/theme-util.c
@@ -95,7 +95,7 @@ gboolean theme_delete (const gchar *name, ThemeType type)
 
     case THEME_TYPE_WINDOW:
       theme = (MateThemeCommonInfo *) mate_theme_info_find (name);
-      theme_dir = g_build_filename (theme->path, "marco-1", NULL);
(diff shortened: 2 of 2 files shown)


lukefromdc:
This works, though I found it was necessary to close and reopen mate-control-center before a deleted theme actually disappeared from the icon view showing the various theme

lukefromdc:
This does not fix the actual problem (m-c-c must be restarted before the deleted theme is removed from the icon view) but specifying a directory that does not exist is still a bug and still should be fixed

zhangxianwei8:
m-c-c do not need to be restarted, it can take effect that theme is removed from the icon view.
Actually, this patch remove the directory metacity-1, such as /usr/share/themes/Menta/metacity-1

raveit65:
> This does not fix the actual problem (m-c-c must be restarted before the deleted theme is removed from the icon view) but specifying a directory that does not exist is still a bug and still should be fixed

This is normal behaviour because the thumbnail needs to be regenerated, which happens only at start of mate-appearance-properties, imo.
So this isn't related to PR, but let me test this for my own.

raveit65:
Is the second commit tested with a theme-installer file?

zhangxianwei8:
Yes, I have tested. You can compress a theme by yourself, such as:
cd /usr/share/themes
tar zcvf Menta.tar.gz Menta
Then, remove "Menta" theme, and then choose the Menta.tar.gz to install with m-a-p. @raveit65

raveit65:
Theme installer is working here too,

raveit65:
merged
https://github.com/mate-desktop/mate-control-center/commit/ce58756f197dcf725338a3ead8a8a93a3633d073
https://github.com/mate-desktop/mate-control-center/commit/173a0e7236c1ea5ea42be1ed91c11d94001e3124

Thank you

monsta:
Nice catch. Probably another case of excessive renaming back in 2011 after the fork.

PR #325, opened 2018-01-22
Add support for HiDPI displays

vkareh:
When setting `GDK_SCALE=2` and `GDK_DPI_SCALE=0.5` as env variables, m-c-c will scale font-rendering DPI values in the Appearance > Fonts > Details app.

This is part of multiple pull requests:
* mate-desktop/mate-settings-daemon/pull/208
* mate-desktop/mate-control-center/pull/325
* mate-desktop/mate-panel/pull/710
* ubuntu-mate/mate-hud/pull/18
* ubuntu-mate/mate-tweak/pull/37


monsta:
Hmm I see 96 there, did a setting reset, still 96. I have 1280x1024 too.

sc0w:
I tested the value before and later:

before: dpi=96.260600
later: dpi=97.000000

vkareh:
@sc0w - the `ceil` is not really necessary, I guess. The dpi spinner only supports integers, so I did that - but you're right, I should just leave it as a full float value in case the user is using the auto-detected value.

@monsta - so this app is strange in that once you set a value, it locks it in. The default value is `0`, which then triggers the auto-detect. But as soon as you play with the spinner, it sets the value in gsettings. You'd need to revert it back to `0` there and re-test.

sc0w:
I have tested, and without **ceil** it works the same like before here

I suggest that change

vkareh:
@sc0w - done! I've also updated mate-settings-daemon with the same change: https://github.com/mate-desktop/mate-settings-daemon/pull/208/commits/45f6d85ccff98684377d83a1aeda9c9e11239e41

muktupavels:
Are you sure that you want expose scaled value to user? Then this saved value at some point will be written to GSettings. What will happen when user changes his monitor back to non-HiDPI?

And what about users that have for some reason dpi already written to GSettings and user has decided to replace his monitor? His DPI will be unscaled...

lukefromdc:
When the user changes their monitor to a small one, they simply need to rescale back down. This at the moment is a manual rather than an autodetected process but it seems to work. I do think the location in the control center is rather buried and ideally would be linked to in a new tab of mate-appearance-properties or even a new "dpi scaling" toplevel button button made by the same file and bringing up the same slider.

raveit65:
Why merging all so fast?
I would prefere not to use env vars if it is possible.

lukefromdc:
I only merged the panel because the other two were already merged, I was keeping a "dev-all" versioned packaged locally until that was done

raveit65:
Any ways, i tested it several days with a non-HIDPI monitor, and it seems to be OK.
Only a gsettings key is missing to enable HIDPI.
But this can be done in a point release, ideally before ubuntu final freeze.
Hope we find a solution for issues with lockscreen and mate-screenshot before reease if HIDPI is enabled.

PR #327, opened 2018-01-26
avoid deprecated GtkStock

sc0w:
The expected with this PR: the same icons like before

Note: I think the translations are affected, so, this is not for 1.20


lukefromdc:
distcheck failed, not sure whether this work or the hidpi commits broke it or how long its been broken though

raveit65:
Distcheck is working for me with current master and  sc0w-master branches.
Using
./autogen.sh && make && make distcheck
I didn't test more as it is for after 1.20 release.

lukefromdc:
I had rebased sc0w-master onto master, maybe that's where the problem came in. Not a priority if this waits for after 1.20 anyway.I just set distcheck running on a couple test builds last night.

PR #329, opened 2018-02-01
Scale font DPI for HiDPI displays

vkareh:
Automatically scales up the font DPI when loading the GtkSpinner and
down-scale when saving it. The result is that the user gets the correct
DPI showing in the Appearance > Fonts > Details app, while gsettings
stores the unscaled DPI.

Related:
* https://github.com/mate-desktop/mate-desktop/pull/302
* https://github.com/mate-desktop/mate-settings-daemon/pull/210
* https://github.com/ubuntu-mate/mate-tweak/pull/39

### DIFF
--- a/capplets/appearance/appearance-font.c
+++ b/capplets/appearance/appearance-font.c
@@ -473,20 +473,20 @@ dpi_load (GSettings     *settings,
 }
 
 static void
-dpi_changed_cb (AppearanceData *data)
+dpi_changed (GSettings *settings,
+	     gchar     *key,
+	     gpointer   user_data)
+{
+  dpi_load (settings, user_data);
+}
+
+static void
+monitors_changed (GdkScreen      *screen,
+		  AppearanceData *data)
 {
   GtkWidget *widget;
   widget = appearance_capplet_get_widget (data, "dpi_spinner");
   dpi_load (data->font_settings, GTK_SPIN_BUTTON (widget));
-}
-
-static void
-dpi_changed (GSettings *settings,
-	     gchar     *key,
-	     gpointer   data)
-{
-  /* Let gsettings catch up to Xft/DPI */
-  g_timeout_add_seconds (1, (GSourceFunc)dpi_changed_cb, data);
 }
 
 static void
@@ -552,9 +552,10 @@ cb_show_details (GtkWidget *button,
     g_signal_connect (widget, "value_changed",
 		      G_CALLBACK (dpi_value_changed), data->font_settings);
 
-    g_signal_connect (data->font_settings, "changed::" FONT_DPI_KEY, G_CALLBACK (dpi_changed), data);
+    g_signal_connect (data->font_settings, "changed::" FONT_DPI_KEY, G_CALLBACK (dpi_changed), widget);
+
     /* Update font DPI when window scaling factor is changed */
-    g_signal_connect (data->interface_settings, "changed::" SCALING_FACTOR_KEY, G_CALLBACK (dpi_changed), data);
+    g_signal_connect (gdk_screen_get_default (), "monitors-changed", G_CALLBACK (monitors_changed), data);
 
     setup_font_sample (appearance_capplet_get_widget (data, "antialias_none_sample"),      ANTIALIAS_NONE,      HINT_SLIGHT);
     setup_font_sample (appearance_capplet_get_widget (data, "antialias_grayscale_sample"), ANTIALIAS_GRAYSCALE, HINT_SLIGHT);
--- a/capplets/appearance/appearance-font.c
+++ b/capplets/appearance/appearance-font.c
@@ -422,7 +422,6 @@ get_dpi_from_x_server (void)
 {
   GdkScreen  *screen;
   double dpi;
-  gint scale;
 
   screen = gdk_screen_get_default ();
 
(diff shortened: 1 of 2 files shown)


capplets/appearance/appearance-font.c
```diff
@@ -458,28 +453,40 @@ static void
...
+
 static void
 dpi_changed (GSettings *settings,
 	     gchar     *key,
-	     gpointer   user_data)
+	     gpointer   data)
 {
-  dpi_load (settings, user_data);
+  /* Let gsettings catch up to Xft/DPI */
+  g_timeout_add_seconds (1, (GSourceFunc)dpi_changed_cb, data);
```
muktupavels: Why this?

muktupavels:
I would use `unscaled` instead of `down-scaled` in commit message.

capplets/appearance/appearance-font.c
vkareh: Since the DPI gets updated in the settings-daemon after the scaling factor is changed, we need to let the settings-daemon process the value and update Xft/DPI before we read this... otherwise, it happens before m-s-d has had a chance to recalculate the DPI and the Appearance app gets the old value.

vkareh: Remember that we're now storing the unscaled DPI, so the control center doesn't know that the DPI changed when we change the scale factor

muktupavels: Yes, but dpi does not change in gsettings when you change scaling factor, no? and if you change dpi in gsetting it is unlikely that scaling factor changed at same time...

I think you want size-changed / monitors-changed signal on GdkScreen to know when scaling factor has changed to update dpi you show in gui.

vkareh: the `size-changed`/`monitors-changed` signal would still require the timeout, the m-c-c would process the signal immediately and reload the old DPI - all the while m-s-d still hasn't calculated the new DPI/Scale

vkareh: I guess the solution may then be to check for changes in `Xft/DPI` directly, as that's the true DPI value being displayed...?

muktupavels: What dpi msd calculates? it sees unscaled value and should pass only scaled value to gtk+.  mcc will get size-changed / monitors-changed signal only after gtk+ will have received updated things, no?

vkareh: wait, you might be right there - why was I seeing the old value then?
...
Ah, okay - I think I know! When changing the scale-factor, m-s-d will update `Gdk/WindowScalingFactor`, which will in turn let GTK scale things up - if I read the `gdk_window_get_scale_factor()` too soon, it still returns the old value, which means I'm multiplying the DPI by the old scale value.

By that logic, I need to monitor for Gdk/WindowScalingFactor so that I can update the GtkSpinButton... `size-changed` shouldn't work, as that in itself triggers a re-calc of the `Gdk/WindowScalingFactor` inside m-s-d (which brings us to the same issue: we would still need the timeout to let m-s-d do its thing).

Nevertheless, I'll try listening in to `size-changed` or `monitors-changed` and see if that works, otherwise I'll have to find a way to listen to changes in `Gdk/WindowScalingFactor`

muktupavels: in msd you know correct values, both dpi and scale factor, no? dpi is always unscaled from gsettings or autocalculated, same with scaling factor - gsettings and callucated. so why would you need size-changed there?

in mcc you just need to update ui when gsettings value (dpi) is changed externally and/or you get monitors-changed / size-changed signal from gtk+. don't remember when which signal is emitted.

vkareh: for msd, the scaling factor can be set to 0 for auto-detect. This means that if a user changes the screen resolution, it should re-calculate it. Hence why I'm listening to `size-changed`.

for mcc, if I understand what you're saying, I should just query for `size-changed` and then look at the monitor scale value? Wouldn't that value be the old one, up until Gdk/WindowScaleFactor is updated? Otherwise I would need to have the same `get_window_scale()` function in mcc that I already have in msd - that wouldn't necessarily be a problem I guess, just seems weird that there would be two places doing their own independent calculation...

muktupavels: You are doing too much in `screen_callback`. You only should check if scaling factor changes and if it is same then it should do nothing. so recalculate window scale and if it is same return. otherwise update setting.

On simple resolution and/or monitor attach / removal it will probably be same value. But I think that at least one of that signals are emitted when window scale has changed. So you might get one signal because of resolution change and then one for scaling factor change. Or am I wrong and window scale change does not trigger any of both signals?

muktupavels: https://git.gnome.org/browse/gtk+/tree/gdk/x11/gdkscreen-x11.c?h=gtk-3-22#n947

Updating window scale calls `process_monitors_change` that can emit `monitors-changed` signal. Check `init_randr15`, geometry depends on scale and if something has changed signal should be emitted.

vkareh:
@lukefromdc - if you change your resolution, but based on the auto-detect it still comes up as scale=1, nothing should happen. That's the correct behavior :)

@muktupavels - thanks for following through on this feedback with me, I tested this and using `"monitors-changed"` worked really well! :) I've updated this PR

monsta:
I think these commits can be squashed before merging.

raveit65:
squashed and merged.
https://github.com/mate-desktop/mate-control-center/commit/856149cbdfafabeb41598b6c818074fca6c0ee1c
Thanks

PR #330, opened 2018-02-11
Remove useless parameter in mate_theme_install_real

zhangxianwei8:
filetype is not used in mate_theme_install_real, remove it.


raveit65:
Thanks.

PR #331, opened 2018-02-11
capplets/mouse: New doubleclick artwork with mate logo

sc0w:
what do you think?

do you like this change?


raveit65:
Not sure if this is really nicer, but the background should be transparent, imo
In result you see the theme bg color.

monsta:
I'd like to keep the light bulb. I think it's a better feedback. 🙂

flexiondotorg:
I agree that the light bulb is better feedback.

sc0w:
my intention with this PR is avoid the gnome background, and the light bulb behavior is not clear for me

If the new user does double click quickly, he see the image double-click-maybe.png, and I see the bulb on, why the user must think the double click doesn't work?

anyway, no problem, I will close this

:-)

raveit65:
Maybe using a animated gif like this?
graphics-einstein-203680

raveit65:
I mean splitting this in 3 images.
1. empty table
2 table with E=
3. table with E=MC^

sc0w:
for me, that is better than the light bulb

we can discuss here the ideas

:)

lukefromdc:
The table with E=MC^ would be really confusing to non-scientific new users, I think that's not the way to go. I vote for keeping the light bulb.

raveit65:
I thought every computer user knows Einstein, but maybe not....

... a good idea for replacing the gnome branding is welcome.

sc0w:
for me, my PR is the best choice, of course, the images can be improved

the mate logo is the first thing the new mate user knows

in the light bulb, the images _double-click-maybe.png_ and _double-click-on.png_ have the bulb on

if double click doesn't work, the user see _double-click-maybe.png_ and the bulb is on

vkareh:
I personally like the lightbulb, but I agree that it's confusing regarding single vs double click. The mate logo, although a really good idea, doesn't necessarily improve upon that. For example, an Ubuntu MATE user will not be as familiar with the core MATE logo either, so there is no clear expectation as to what each state should look like.

We need something that has three states naturally, or that the second step makes is clear that there is yet another one...

PR #333, opened 2018-02-17
theme-thumbnail.c: avoid deprecated GtkButton:use-stock

sc0w:


lukefromdc:
I found unrelated issue https://github.com/mate-desktop/mate-control-center/issues/334
during this test, it is NOT AFFECTED by this PR

PR #337, opened 2018-02-27
Fix screen geometry to account for rotation when snapping.

### LINKED ISSUE #198: mate-display-properties: unexpected behaviour
The GUI always belevies the physical width or height of a display are the same regardless of user selected rotation. If you have more than one display and rotate them left or right you cannot place them logically edge to edge in the GUI since it beleives it will overlap.  I manually edited the monitors.xml to fix this. Perhaps the GUI should have borth physical properties (HW layer) and logical properties, so it doesent get confused.
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26550533-mate-display-properties-unexpected-behaviour?utm_campaign=plugin&utm_content=tracker%2F559197&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F559197&utm_medium=issues&utm_source=github).
</bountysource-plugin>

BrentBradburn:
resolves #198 

The applet snaps to edges, but chooses the edge locations as if no displays are rotated. This patch fixes that by reading the rotated geometry where necessary.

### DIFF
--- a/capplets/display/xrandr-capplet.c
+++ b/capplets/display/xrandr-capplet.c
@@ -1153,6 +1153,7 @@ list_edges_for_output (MateRROutputInfo *output, GArray *edges)
     int x, y, w, h;
 
     mate_rr_output_info_get_geometry (output, &x, &y, &w, &h);
+    get_geometry(output, &w, &h); // accounts for rotation
 
     /* Top, Bottom, Left, Right */
     add_edge (output, x, y, x + w, y, edges);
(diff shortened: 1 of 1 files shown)


raveit65:
Calling Ubuntu maintainer.

lukefromdc:
There is a momentary delay before redrawing the desktop which switching from window-scaling = 2 to window-scaling = 1 as well, only a couple seconds.

monsta:
@flexiondotorg @vkareh 
It's a fix for https://bugs.launchpad.net/bugs/1576684 - are you guys able to test this?

lukefromdc:
The redraw issue I saw was very short duration and quickly self-correcting. As for the orignal issue, none of my monitors have the mounting hardware to be set up in portrait mode, and I found out years ago that I cannot control the mouse on a display rotated off of its intended mode.

monsta:
Ok, I guess this is good to go, as the mentioned issues aren't caused by this PR.

PR #338, opened 2018-03-15
Add auto-detect toggle switch

vkareh:
Currently if a user changes the DPI, there is no way to reset it back to the auto-detected value from Xserver (which we store as 0 in gsettings). Adding this toggle solves that issue.

I also removed a deprecated [GtkHandleBox](https://developer.gnome.org/gtk3/stable/GtkHandleBox.html) and all its associated code.

### DIFF
--- a/capplets/appearance/appearance-font.c
+++ b/capplets/appearance/appearance-font.c
@@ -473,11 +473,22 @@ dpi_load (GSettings     *settings,
 }
 
 static void
-dpi_changed (GSettings *settings,
-	     gchar     *key,
-	     gpointer   user_data)
+dpi_changed (GSettings      *settings,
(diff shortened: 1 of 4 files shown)


vkareh:
@lukefromdc - Thanks for testing. The switch is a nice change from checkbox. My original attempt had a checkbox and I could not make it look nice in any position. The switch just seemed to flow :slightly_smiling_face:

lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/mate-control-center/commit/48b4fc3407816ddcab6e2bd20b063b6de9746e25

sc0w:
I have tested now, and the behavior is the expected

thanks @vkareh

PR #340, opened 2018-03-24
Scale cursor size for HiDPI when changing themes

vkareh:
Related: mate-desktop/marco/pull/400

### DIFF
--- a/capplets/common/mate-theme-info.c
+++ b/capplets/common/mate-theme-info.c
@@ -32,6 +32,7 @@
 #include <glib/gi18n.h>
 #include <gmodule.h>
 #include <gtk/gtk.h>
+#include <gdk/gdk.h>
 #include <gdk/gdkx.h>
 #include <gio/gio.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


raveit65:
Some minor issues.
1. If i switch to scale=2 and mate-appearance-properties was open, switching the cursor theme has issues.
Works well if i open m-a-p after switching to scale=2.
Imo, we can can ignore that because 'life isn't a cherry bowle' ;-)

2. If i select our ContrastHighInverse theme in m-a-p, i need to click 2 times on the thumbnail before i see the right cursor size. This i can fix with defining a cursor theme in index.theme file.
https://github.com/mate-desktop/mate-themes/blob/master/desktop-themes/ContrastHighInverse/index.theme.in
Eg.
```
[X-GNOME-Metatheme]
GtkTheme=ContrastHighInverse
IconTheme=ContrastHigh
MetacityTheme=ContrastHighInverse
CursorTheme=mate
```

3. The slider for setting the cursor size doesn't work with most cursor themes with HIDPI,
except adwaita-cursor-theme.
See.
https://git.gnome.org/browse/adwaita-icon-theme/commit/?id=7c74b771842536c746a200fc4093818a549c1a2f
and explanation
https://bugzilla.gnome.org/show_bug.cgi?id=747386
But i have no idea about the structure of a image/x-cursor and how i can edit them. 
Would be nice if this can be fixed without editing the cursor theme.

Any ways, i am +1 for merging this, as using the cursor is fixed in most cases and we need a 1.20 point release the next days.

PR #342, opened 2018-04-25
Change cursor-size to 24px

vkareh:
The MATE cursor theme does not have an 18px cursor size, so having
a 18px as the default setting does not make sense. Using 24px looks
better in modern displays and allows for proper HiDPI scaling.

Relevant: mate-desktop/mate-desktop/pull/313

### DIFF
--- a/capplets/appearance/appearance-style.c
+++ b/capplets/appearance/appearance-style.c
@@ -544,7 +544,7 @@ update_cursor_size_scale (MateThemeCursorInfo *theme,
     if (theme && theme->sizes->len > 0)
       size = g_array_index (theme->sizes, gint, 0);
     else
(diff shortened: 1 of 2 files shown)


lukefromdc:
I cannot really review this one due to system issues that leave my normal cursor impossible to change. This is almost certainly a leftover from steps I took a while ago to keep KDE's Breeze theme from changing the default cursor theme but now I can't find it

raveit65:
@sc0w 
Do you agree to change cursor-size for Trad themes too?

sc0w:
@raveit65 yes, I have no problem

lukefromdc:
Merging as all three reviewes here have approved.

PR #344, opened 2018-04-27
Support mate-menus api change

yetist:
Please review and see this [PR](https://github.com/mate-desktop/mate-menus/pull/59).

### DIFF
--- a/libslab/app-shell.c
+++ b/libslab/app-shell.c
@@ -871,8 +871,9 @@ void
 generate_categories (AppShellData * app_data)
 {
 	MateMenuTreeDirectory *root_dir;
-	GSList *contents, *l;
 	gboolean need_misc = FALSE;
+	MateMenuTreeIter *iter;
+	MateMenuTreeItemType type;
 
 	if (!app_data->tree)
 	{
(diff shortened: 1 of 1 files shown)


yetist:
Well, I think mozo should change it's desktop file such as category and keywords, when done that, it will show in the control center maybe.

lukefromdc:
I had Mozo showing up fine in the control center after I rebuilt it from the PR version though it and Alacarte look almost exactly alike in my setup and can be hard to tell apart.

Proof that it showed up is this: When I removed my packaged version and built and installed from the PR, I forgot to recompile the gtk icon cache and it showed up with the missing-icon icon. After running
gtk-update-icon-cache -q -t -f /usr/share/icons/hicolor
the icon for mozo showed up normally. I tested these as I built them, first mate-menus, then the panel, then the control center, and finally mozo. All four need to be merged together and used together. The version bump is this important, in the case of the panel we have already bumped to 1.21.0 and thus might want to bump again to 1.21.1 . There would be no problem at final release though as they would all be 1.22 versions and I don't know of any distro building development versions anymore.

raveit65:
Version bumping should be done after merging by us, not in PRs.
Because first we need to to release and tag mate-menus before we can increase requirements.
Honestly i am missing some statements from all members of core team in mate-menus PR about re-forking mate-menus.

lukefromdc:
I agree that the version bump comes afterwards, especially as we might want to add other things and it might involve translations that must be done first. Easy enough to roll it back in the PR's, then bump it in Master later and in sync.

monsta:
> Version bumping should be done after merging by us, not in PRs.

Actually, https://github.com/mate-desktop/mate-menus/pull/59 has version bump. Do we need it?

lukefromdc:
We do in fact need to bump the versions here, as the API changes and the new and old versions cannot be mixed. Needs to be done prior to these commits appearing in any release but not necessarily in Master

monsta:
We'll need it before release for sure - the question was whether to bump versions after merge or to allow it in PR itself.

libslab/app-shell.c
```diff
@@ -888,42 +889,35 @@ generate_categories (AppShellData * app_data)
...
+				item = matemenu_tree_iter_get_directory (iter);
+				category = matemenu_tree_directory_get_name (item);
+				generate_category(category, item, app_data, TRUE);
+				matemenu_tree_item_unref (item);
+				break;
+			case MATEMENU_TREE_ITEM_ENTRY:
+				need_misc = TRUE;
+				break;
+			default:
+				g_assert_not_reached ();
```
monsta: `MateMenuTreeItemType` enum has more possible values than these two, so why introducing the assertion? Failed assertion means immediate app abort.

libslab/app-shell.c
```diff
@@ -1022,52 +1025,54 @@ generate_launchers (MateMenuTreeDirectory * root_dir, AppShellData * app_data, C
...
+							desktop_file);
+					break;
+				}
+				if (!check_specific_apps_hack (desktop_item))
+					insert_launcher_into_category (cat_data, desktop_item, app_data);
+				mate_desktop_item_unref (desktop_item);
+				matemenu_tree_item_unref (item);
+				break;
+			default:
+				g_assert_not_reached ();
```
monsta: Same here, similar assert

raveit65:
@monsta
Is this ready to go?

monsta:
This one looks ok now, unless I missed something.

raveit65:
@monsta 
If it looks Ok please approve. Than everybody know that all is fine :)

raveit65:
A bit squashed and merge.
https://github.com/mate-desktop/mate-control-center/commit/7ba869d95f3c01323ecd617cdead6f7d03d0ae51
Thanks a lot.

raveit65:
I have released m-c-c-1.21.0 but not tagged yet, available for testing in pre-release folder at server.

Issue #346, opened 2018-05-01
Feature request: Add buttons-have-icons control to Appearance's Interface tab


elcste:
Ubuntu MATE has icons on buttons by default, but a lot of people like the cleaner look of removing them. Of course you can use `org.mate.interface buttons-have-icons false` [edit: fixed key] (and I do), but this is not is not very discoverable or convenient for a lot of people.

I propose adding this option to the Interface tab of Appearance settings. This would mirror the very similar setting for showing icons on menus that has a control there.

#### Expected behaviour

Have an option in the GUI to control the buttons-have-icons setting so we don't have to use gsettings to change this option.

#### Actual behaviour

No GUI option exists AFAICT. This used to be available in MATE Tweak's Interface section, but that section was removed because the other options are no longer needed.

#### Steps to reproduce the behaviour

N/A

#### MATE general version

1.20.1

#### Package version

1.20.2-2ubuntu1

#### Linux Distribution

Ubuntu MATE Ubuntu 18.04

#### Link to downstream report of your Distribution

https://bugs.launchpad.net/ubuntu/+source/mate-control-center/+bug/1768165

lukefromdc:
You can turn them off in dconf-editor, which can handle most of the gsettings options (I think all of them including the ones in relocatable schemas in the newest versions).

elcste:
True, but I still think it's less discoverable or convenient for a lot of people, and the same kind of option for menu icons is in the settings ;-)

Thanks for your consideration.

alexarnaud:
Le 01/05/2018 à 03:45, lukefromdc a écrit :
> You can turn them off in dconf-editor, which can handle most of the 
> gsettings options (I think all of them including the ones in relocatable 
> schemas in the newest versions).

On the Mate Appearance Property, here Mate 1.16 on Debian, there is an 
interface tab where I can choose to remove icon on menu items.

We could add similar feature for buttons so :).

Best regards,
Alex.

raveit65:
> Of course you can use `gsettings set org.gnome.desktop.interface buttons-have-icons false`

... I am in doubt that this key does have any effect in MATE session :)
```
[rave@mother ~]$ gsettings set org.gnome.desktop.interface buttons-have-icons false
No such key ?buttons-have-icons?
```

Btw. new GUI of dconf-editor looks much better with
dconf-editor-3.28.0-1.fc28.x86_64

elcste:
[...]

PR #352, opened 2018-05-16
capplets/appearance: add buttons-have-icons switch

ericek111:
Requested in https://github.com/mate-desktop/mate-control-center/issues/346

Handles `org.mate.interface buttons-have-icons`

### DIFF
--- a/capplets/appearance/appearance-ui.c
+++ b/capplets/appearance/appearance-ui.c
@@ -25,55 +25,57 @@
 static void
 set_have_icons (AppearanceData *data, gboolean value)
 {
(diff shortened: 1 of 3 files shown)


ericek111:
:( I've fixed the formatting, it was a mess of spaces and tabs, that confused me. Is it okay now?
Thank you.

raveit65:
I ask only for editing the part which you had already edited. :)
But OK, it looks good now, thx for doing this.
We can squash/rebase the both indents commit in one during merge.
A separate commit for the indents fix is better as to have all in one commit.
To put all In one commit makes it hard to see the real commit later.

PR #356, opened 2018-06-08
libslab: code clean up

raveit65:
Since we use GtkSearchEntry a lot of code can be dropped.
https://github.com/mate-desktop/mate-control-center/commit/afcb324

Eg: GtkSearchEntry has a own menu, in result search-context-picker.c
does nothing.
So other code in searchbar.c isn't used.

Hope i didn't break something.
Please test.


libslab/search-bar.h
```diff
@@ -43,7 +43,7 @@ typedef struct
 {
 	GtkBoxClass parent_class;
 
-	void (*search) (NldSearchBar *, int context_id, const char *text);
+	void (*search) (NldSearchBar *, const char *text);
```
muktupavels: I think you should remove `context_id` parameter also from signal...

raveit65: Where do you see the occurrence of context_id ?
grep gives me no result...

muktupavels: https://github.com/mate-desktop/mate-control-center/blob/3b621c82730fa216e7b4d19e9f605d99953823b1/libslab/search-bar.c#L73

raveit65: This place i know, i don't see context_id here...

muktupavels: You have removed `context_id` parameter from virtual function. Quick look shows that this is public library - so this is API/ABI break...

This function is used with signal - `G_STRUCT_OFFSET (NldSearchBarClass, search)` and signal still have two parameters - `2, G_TYPE_INT, G_TYPE_STRING`. First is `context_id` and second is `text`. If I remember correctly then virtual function should have same parameters as signal.

If search is not `NULL` (is is here, but might might not be in subclass in other app that uses this library) then it will be called when signal is emitted. Now imagine what will happen when glib will try to call your search function `(NldSearchBar *, const char *text)` with `(search_bar, some_id, some_text)`. What will happen when you will try to access/use `text`?

raveit65: Removing `G_TYPE_INT,` is a bad idea. With
```
	signals[SEARCH] =
		g_signal_new ("search", G_TYPE_FROM_CLASS (nld_search_bar_class),
		G_SIGNAL_RUN_FIRST | G_SIGNAL_ACTION, G_STRUCT_OFFSET (NldSearchBarClass, search),
		NULL, NULL, nld_marshal_VOID__INT_STRING, G_TYPE_NONE, 2, G_TYPE_STRING);
```
m-c-c crashes if i type something in search field or i get this error message.
```
(mate-control-center:16403): GLib-GObject-CRITICAL **: 08:51:41.190: g_value_type_compatible: assertion 'G_TYPE_IS_VALUE (src_type)' failed

(mate-control-center:16403): GLib-GObject-WARNING **: 08:51:41.190: gsignal.c:3375: invalid object type '
```

muktupavels: Of course it is bad idea, if you don't understand how things works...

First your existing changes are already bad and could lead to crashes, but that depends if there is any code / app that uses this library outside mcc, I don't know that...

There might be something like this:
```
...
#define MY_TYPE_SEARCH_BAR (my_search_bar_get_type ())
G_DECLARE_FINAL_TYPE (MySearchBar, my_search_bar, MY, SEARCH_BAR, NldSearchBar)
...

...
struct _MySearchBar
{
  NldSearchBar parent;
};

G_DEFINE_TYPE (MySearchBar, my_search_bar, NLD_TYPE_SEARCH_BAR)
...

static void
my_search_bar_search (NldSearchBar *search_bar,
                      int           context_id,
                      const char   *text)
{
  g_warning ("search: context_id - %d, text - %s", context_id, text);
}

static void
my_search_bar_class_init (MySearchBarClass *bar_class)
{
  bar_class->search = my_search_bar_search;
}
...
```

Your change would mean that this code will not compile without changes and also would crash with new library if it is not recompiled.

Safest thing would be to not remove `context_id` if you don't know how this library is / could be used outside mcc, but...

Is this library used by other applications? Debian code search seems to tell that it is used only by mcc. If so then why this library should be public / installed?

If you know that mcc is only user of this library then you can remove that virtual function completely and replace `G_STRUCT_OFFSET (NldSearchBarClass, search)` with `0`.

Now about your attempt at removing `G_TYPE_INT`... Did you pay attention to `2, G_TYPE_INT, G_TYPE_STRING` part in my comment? There was reason why `2` was included. I will let you find what that number means in documentation...

Also just removing parameter there is not enough and worst part here is that compiler will not warn you. You need to remove that parameter from `g_signal_emit` (currently you have replaced `nld_search_bar_get_context_id` with `0`) and also everywhere you connect to `search` signal. One such thing is in `app-shell.c`...

raveit65: > First your existing changes are already bad and could lead to crashes, but that depends if there is any code / app that uses this library outside mcc, I don't know that...

Why you post that if you are not sure?
This libary isn't used outside of libslab. Why it should?

> There was reason why 2 was included. I will let you find what that number means in documentation...

Please tell me what i should do. Sorry, i don't have the time to play games here...

muktupavels: > Why you post that if you are not sure?

Because I see that this library is installed for external use and this change is API and ABI break. Because virtual function that is used with signal should have same parameters?

> This libary isn't used outside of libslab. Why it should?

Because you provide this library and someone has decided that it might be good library to use for specific task or something like that... You know, I can install `libmate-slab-dev` and start using in my app anytime I want! If that is not supposed to be used outside of mcc then start cleanup by turning this library in non install library...

> Please tell me what i should do. Sorry, i don't have the time to play games here...

It is your fix / your change and you should learn what code is doing or how it is working... Open glib documentation and read what parameters `g_signal_new` has.

If you don't have time then go ahead and merge this without any further changes. After all your teammate approved changes. :)

raveit65: Yes, we have a team :smile:

raveit65: Using
```
	signals[SEARCH] =
		g_signal_new ("search", G_TYPE_FROM_CLASS (nld_search_bar_class),
		G_SIGNAL_RUN_FIRST | G_SIGNAL_ACTION, 0,
		NULL, NULL, nld_marshal_VOID__INT_STRING, G_TYPE_NONE, 1, G_TYPE_STRING);
```
and
```
	g_signal_emit (search_bar, signals[SEARCH], 0,
	               nld_search_bar_get_text (search_bar));
```
gives me this runtime warning when typing
```
** (mate-control-center:15741): CRITICAL **: 16:03:09.153: nld_marshal_VOID__INT_STRING: assertion 'n_param_values == 3' failed
```

muktupavels: In how many places `nld_marshal_VOID__INT_STRING` is used? If this is only place then you could update `nld-marshal.list` and change `VOID:INT,STRING` to `VOID:STRING` or simply add it as new line there, but I think you can replace it with NULL and delete `nld-marshal.list`.

https://developer.gnome.org/gobject/stable/howto-signals.html
> The C signal marshaller should always be NULL, in which case the best marshaller for the given closure type will be chosen by GLib. This may be an internal marshaller specific to the closure type, or g_cclosure_marshal_generic, which implements generic conversion of arrays of parameters to C callback invocations. GLib **used to require** the user to write or generate a type-specific marshaller and pass that, but that has been deprecated in favour of automatic selection of marshallers.

vkareh: @raveit65 - functionally, this seems to work, but I agree with @muktupavels here. Now the signature is mis-matched (the signal callback expects the `int context_id` to be sent, but the signature of the class method object no longer has the `context_id`).

I changed to `VOID:STRING` and `nld_marshal_VOID__STRING`, then adjusted the callback in `app-shell.c` to remove the `context_id` from there and it seems to work fine.

Here's my diff, in case you're interested (I can also commit to your branch, but this is an easy enough change I think):
```diff
diff --git a/libslab/app-shell.c b/libslab/app-shell.c
index ba13fa4..05d5617 100644
--- a/libslab/app-shell.c
+++ b/libslab/app-shell.c
@@ -74,7 +74,7 @@ static void launch_selected_app (AppShellData * app_data);
 static void generate_potential_apps (gpointer catdata, gpointer user_data);
 
 static void relayout_shell (AppShellData * app_data);
-static gboolean handle_filter_changed (NldSearchBar * search_bar, int context, const char *text,
+static gboolean handle_filter_changed (NldSearchBar * search_bar, const char *text,
 	gpointer user_data);
 static void handle_group_clicked (Tile * tile, TileEvent * event, gpointer user_data);
 static void set_state (AppShellData * app_data, GtkWidget * widget);
@@ -581,7 +581,7 @@ handle_filter_changed_delayed (gpointer user_data)
 }
 
 static gboolean
-handle_filter_changed (NldSearchBar * search_bar, int context, const char *text, gpointer data)
+handle_filter_changed (NldSearchBar * search_bar, const char *text, gpointer data)
 {
 	AppShellData *app_data;
 
diff --git a/libslab/nld-marshal.list b/libslab/nld-marshal.list
index 158cf7a..30ba5d8 100644
--- a/libslab/nld-marshal.list
+++ b/libslab/nld-marshal.list
@@ -1 +1 @@
-VOID:INT,STRING
+VOID:STRING
diff --git a/libslab/search-bar.c b/libslab/search-bar.c
index ac28de5..157c4ba 100644
--- a/libslab/search-bar.c
+++ b/libslab/search-bar.c
@@ -73,8 +73,7 @@ static void nld_search_bar_class_init (NldSearchBarClass * nld_search_bar_class)
 	signals[SEARCH] =
 		g_signal_new ("search", G_TYPE_FROM_CLASS (nld_search_bar_class),
 		G_SIGNAL_RUN_FIRST | G_SIGNAL_ACTION, G_STRUCT_OFFSET (NldSearchBarClass, search),
-		NULL, NULL, nld_marshal_VOID__INT_STRING, G_TYPE_NONE, 2, G_TYPE_INT,
-		G_TYPE_STRING);
+		NULL, NULL, nld_marshal_VOID__STRING, G_TYPE_NONE, 1, G_TYPE_STRING);
 }
 
 static void
@@ -166,7 +165,7 @@ emit_search (NldSearchBar * search_bar)
 		priv->timeout_id = 0;
 	}
 
-	g_signal_emit (search_bar, signals[SEARCH], 0, 0,
+	g_signal_emit (search_bar, signals[SEARCH], 0,
 		nld_search_bar_get_text (search_bar));
 }
 ```

No the other issue is API compatibility - I seriously doubt anyone uses this outside of m-c-c, but you never know. Since this is a clean-up, it probably won't be backported to <=1.20 and will be released with 1.22, at which point I'm fine if that one guy that decided to use libslab years ago for their pet project needs to update :stuck_out_tongue_closed_eyes:

Comment (raveit65) [reply libslab/search-bar.h]: wtf, someone makes my life complicated :)
But step by step, let me first commit the signal changes before making it internal.
Btw. none of my RPMS for fedora are depending on mate-control-center-devel.

Comment (raveit65) [reply libslab/search-bar.h]:
```
[root@satellite rave]# dnf repoquery --whatrequires mate-control-center-devel
Last metadata expiration check: 3:24:32 ago on Sat Jun  9 14:37:38 2018.
[root@satellite rave]# 
```

Comment (raveit65) [reply libslab/search-bar.h]: Adding your suggestions like this gives me again a warning
```
static void nld_search_bar_search (NldSearchBar *search_bar,
	                               const char   *text)
{
	g_warning ("search: text - %s", text);
}

static void nld_search_bar_class_init (NldSearchBarClass * nld_search_bar_class)
{
	GObjectClass *object_class = G_OBJECT_CLASS (nld_search_bar_class);
	GtkWidgetClass *widget_class = GTK_WIDGET_CLASS (nld_search_bar_class);

	nld_search_bar_class->search = nld_search_bar_search;

	object_class->finalize = nld_search_bar_finalize;
	widget_class->focus = nld_search_bar_focus;
	widget_class->grab_focus = nld_search_bar_grab_focus;

	g_type_class_add_private (nld_search_bar_class, sizeof (NldSearchBarPrivate));

	signals[SEARCH] =
		g_signal_new ("search", G_TYPE_FROM_CLASS (nld_search_bar_class),
		G_SIGNAL_RUN_FIRST | G_SIGNAL_ACTION, G_STRUCT_OFFSET (NldSearchBarClass, search),
		NULL, NULL, nld_marshal_VOID__STRING, G_TYPE_NONE, 1, G_TYPE_STRING);
}
```
```
** (mate-control-center:8325): WARNING **: 17:56:51.551: search: text - d
```

muktupavels: You were supposed to test that without `g_signal_new` changes to see what could happen when parameters does not match. You don't really need it. :)

raveit65: OK, than i am a lucky boy :)
I will squash the commits before i start to make the libs internal.

PR #368, opened 2018-08-21
accounts service improve

raveit65:
This is the same like https://github.com/mate-desktop/mate-control-center/pull/362
I only added a configure check for accountsservice, to avoid  a build breakage when the package doesn't exists in distros, eg. bsd.

Please test.


lukefromdc:
If someone would tell me how to reset the face picture to blank, I can then test the drag and drop function.

raveit65:
That's a good question.
Honestly i have no idea where accountsservice store this.

vkareh:
@lukefromdc - in the About Me window, click on the picture icon so that the dialog pops up, at the bottom there is a button that says "No Image", which clears the face picture.
screenshot at 2018-08-24 00-59-06

vkareh:
@lukefromdc - https://cgit.freedesktop.org/accountsservice/tree/src/user.c#n133
```c
user->default_icon_file = g_build_filename (home_dir, ".face", NULL);
```
(which means that our code does exactly the same as libaccountservie) :stuck_out_tongue:

vkareh:
So a re-login did not do anything. I still get a segfault, but the underlying cause seems to be that the accountsservice does not believe I'm a user... :interrobang:

lukefromdc:
Now THAT was something I didn't think to test: a face picture that could not be loaded. I agree that must be fixed, as this sort of thing that locks a user out is known for triggering reinstalls of the whole OS by inexperienced users, sometimes to include trashing all data that missed the last (if any...)backup.

vkareh:
Actually the picture doesn't seem to be the problem... it seems to be inside `manage_get_user`, that doesn't find my user. The rest just cascades down and fails...

But it's _way_ past bedtime now, I'll take a look at this tomorrow to see if I can find the root cause.

raveit65:
@yetist 
ping

yetist:
@vkareh 
Please test your AccountService use follow code:
```
#include <act/act.h>
// gcc -o users users.c `pkg-config --cflags --libs accountsservice gobject-2.0`

int main(int argc, char** argv)
{
	ActUserManager* manager;
	const char* name;
	const gchar   *file;
	GSList *lst, *l;

	manager = act_user_manager_get_default ();
	lst = act_user_manager_list_users(manager);
	g_print("All users:\n");
	for(l=lst; l; l=l->next) {
		name = act_user_get_user_name(l->data);
		file = act_user_get_icon_file (l->data);
		g_print("name: %-10s \ticon: %s\n", name, file);
	}
	g_slist_free_full (lst, g_object_unref);
	return 0;
}
```
My output like this:
```
$ ./users 
All users:
name: yetist     	icon: /var/lib/AccountsService/icons/yetist
name: isoft      	icon: /home/isoft/.face
name: xx1        	icon: /home/xxxxx/.face
name: zzzzzzz    	icon: /home/zzzzzzz/.face
```

raveit65:
Good to know that /var/lib/AccountsService/icons/ is main dir.

muktupavels:
Does anyone read documentation?

 `act_user_manager_get_default` is `transfer none` so you probably should not unref it... Seems that **if** you know what you are doing unref-ing should not cause problems.

> Calling this function will automatically being loading the user list if it isn't loaded already. The #ActUserManager:is-loaded property will be set to %TRUE when the users are finished loading and then act_user_manager_list_users() can be called.

To me this sounds like you should call `act_user_manager_list_users` when `is-loaded` property is set to `TRUE`, but code does not check that...

`act_user_manager_list_users` has `transfer container`... In other words - users are owned by user manager and returned objects should not be unref-ed. You should call `g_slist_free` not `g_slist_free_full`.

If I did not miss anything, then `me->user` is invalid right after you unref manager. That would explain:

> act_user_get_icon_file: assertion 'ACT_IS_USER (user)' failed

Segfault happens because `gdk_pixbuf_new_from_file` expects non-NULL filename. In this case it just returns `NULL` **without** setting error.

raveit65:
@yetist 
Do you agree with concerns from @muktupavels  ?
Maybe it's better to work in your PR again?
You can add my commit or we add this later.

yetist:
@raveit65 @muktupavels @vkareh 
#362 has updated, please test, thank you.

raveit65:
Closing this in favour of origin PR.

PR #371, opened 2018-08-31
for accounts service support

yetist:
See [here](https://github.com/mate-desktop/mate-control-center/pull/362#issuecomment-417329043)


capplets/about-me/mate-about-me.c
```diff
@@ -410,12 +410,10 @@ about_me_fingerprint_button_clicked_cb (GtkWidget *button, MateAboutMe *me)
...
 {
-        if (act_user_is_loaded (user)) {
-		about_me_load_photo (me);
-		g_signal_handlers_disconnect_by_func (G_OBJECT (user),
-				G_CALLBACK (on_user_is_loaded_changed),
-				me);
+	if (!act_user_is_loaded (user)) {
+		return;
 	}
+	about_me_load_photo (me);
```
muktupavels: Anyway changes to `on_user_is_loaded_changed` is unneeded, no? It does same, except removed `g_signal_handlers_disconnect_by_func`...

capplets/about-me/mate-about-me.c
```diff
@@ -480,11 +478,14 @@ about_me_setup_dialog (void)
 	g_object_get (manager, "is-loaded", &loaded, NULL);
 	if (!loaded) {
 		g_signal_connect (me->user, "notify::is-loaded", G_CALLBACK (on_user_is_loaded_changed), me);
+	} else {
+		/* Contact Tab */
+		about_me_load_photo (me);
```
muktupavels: This will never happen! `is-loaded` is and will be `FALSE` here.

muktupavels:
> Don't get me wrong - I make mistakes too, so feel free to correct me if I am wrong. :)

Don't blindly copy code that I show as example! For example in this case you could argue that `about_me_load_photo` function call is needed to load initial (avatar-default) image. I would say that some loading indicator could be better, but probably not worth doing it as user is loaded fast enough to not see that.

So better thing would be to do something like this:
```
#if HAVE_ACCOUNTSSERVICE
	const gchar   *act_file;

	if (act_user_is_loaded (me->user)) {
		act_file = act_user_get_icon_file (me->user);
		if ( act_file != NULL && strlen (act_file) > 1) {
			file = g_strdup (act_file);
		}
	}
#endif
```

So I would do above change and remove `g_object_get` part. That way `avatar-default` will be used as initial image and once user is loaded it will be updated to use user image.

raveit65:
@yetist 
Is this Ok for you?
For your info @muktupavels is gnome-mantainer (metacity, gnome-panel, and others) from gnome-flashback desktop and he likes to help out with reviews.

yetist:
@muktupavels @raveit65 
Now rebase to upstream master,  is that ok?

raveit65:
@yetist 
Cool that you are back again, and yes `git rebase master` on your branch would be nice.

PR #372, opened 2018-08-31
Fixed image size

yetist:
Fixed EImageChooser's bug for mate-about-me, now it show fixed size image.

This bug can be found through testing:

For example, if you set different images one after another as user face, even if these images have the same size, they will still cause the displayed images to grow larger.

The follow image use the same file /usr/share/pixmaps/faces/book.jpg, but after change the user face  many times, it will show lager:

On the mate-about-me dialog, I think we should always show a fixed size picture.

### DIFF
--- a/capplets/about-me/e-image-chooser.c
+++ b/capplets/about-me/e-image-chooser.c
@@ -29,17 +29,18 @@
 
 #include "e-image-chooser.h"
 
+typedef struct _EImageChooserPrivate EImageChooserPrivate;
 struct _EImageChooserPrivate {
 
 	GtkWidget *image;
-	GtkWidget *browse_button;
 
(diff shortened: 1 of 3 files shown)


muktupavels:
Not going to review code, but what bug? Is there any report? Might be good idea to provide more info about bug and why using fixed size is right fix. If for whatever reason in future some developer will look why fixed size was introduced it will help if there is good info...

lukefromdc:
What is this supposed to fix?

yetist:
Not ready to write the new issue.
A detailed description of this issue has been updated in RP above.

lukefromdc:
OK, I can confirm the bug. Note that this takes place very gradually but does happen. About to test the code

PR #383, opened 2018-11-06
mate-about-me: interact with passwd via /dev/ptmx

### LINKED ISSUE #384: mate-about-me can't change password on OpenIndiana
mate-about-me dialog hangs when we try to change password on OpenIndiana. The issue is that /usr/bin/passwd on illumos asks pam for new user password, and pam opens tty to read it. So we either have "Enter password:" prompt on /dev/tty or passwd dies if /dev/tty is not present. In any case this is not what mate-about-me waits and mate-about-me hangs. 

Corresponding OpenIndiana issue: https://www.illumos.org/issues/9957
Suggested fix: #383 

I'd be glad if people on Linux/FreeBSD checked if it works for them in current form. If not, we can either fix it or make function dependent on ```#ifdef __sun```.

pyhalov:
Fixes #384

### DIFF
--- a/capplets/about-me/mate-about-me-password.c
+++ b/capplets/about-me/mate-about-me-password.c
@@ -649,6 +649,11 @@ io_watch_stdout (GIOChannel *source, GIOCondition condition, PasswordDialog *pdi
 							  "match",
 							  "1 numeric or special",
 							  "failure",
+							  "must contain",
+							  "must differ",
+							  "are not allowed",
+							  "repeating characters",
+							  "history",
 							  "rotated",
 							  "error",
 							  "BAD PASSWORD",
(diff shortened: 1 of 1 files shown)


pyhalov:
@lukefromdc tried to follow linux man pages, however don't have a system to check compilation

lukefromdc:
Got another build error, this time
```

mate-about-me-password.c: In function ‘spawn_passwd’:
mate-about-me-password.c:240:28: error: ‘TIOCSCTTY’ undeclared (first use in this function)
   if (ioctl(pdialog->pty_s,TIOCSCTTY, NULL) < 0) {

```

lukefromdc:
Got another build error, maybe from a typo ?

```
mate-about-me-password.c: In function ‘spawn_passwd’:
mate-about-me-password.c:239:6: warning: implicit declaration of function ‘pipe2’; did you mean ‘pipe’? [-Wimplicit-function-declaration]
  if (pipe2(err_pipe, O_CLOEXEC) < 0) {
      ^~~~~
      pipe
mate-about-me-password.c:239:22: error: ‘O_CLOEXEC’ undeclared (first use in this function); did you mean ‘FD_CLOEXEC’?
  if (pipe2(err_pipe, O_CLOEXEC) < 0) {
                      ^~~~~~~~~
                      FD_CLOEXEC

```

pyhalov:
ptsname(3C) on OpenIndiana says:
```Upon successful completion, the function ptsname() returns a pointer to
       a string which is the name of the pseudo-terminal slave device. This
       value points to a static data area that is overwritten by each call to
       ptsname(). Upon failure, ptsname() returns NULL.```
ptsname(3) on Linux says:
```On success, ptsname() returns a pointer to a string in static storage which will be overwritten by subsequent calls.  This pointer must not be freed.  On failure, NULL is returned```

lukefromdc:
OK, that part is probably the same everywhere, good to to

pyhalov:
Note, that last change (s/return/exit/) will fail unless #387 is also applied. I was a bit puzzled, why we can't exit() from child on error, when we should (we don't want to run duplicate process). Now we know the reason :)

lukefromdc:
Any more updates on this?

lukefromdc (commented):
Builds clean on Debian. Tried for a functional test but I've got some other issue preventing the capplet from recognizing the sudo passphrase both in master and with this applied. I can copy the passphrase into the text buffer, use it to mount devices from caja. change CPUfreq applet settings, use it in terminal etc but authentication fails in this capplet. Same problem with master so not related to this PR.

I got this warning both in master and with this appled:
`
(mate-about-me:21182): capplet-common-WARNING **: 14:33:15.376: Could not get default fprint device: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name net.reactivated.Fprint was not provided by any .service files`

Issue #390, opened 2018-11-14
Keybindings: Cannot cancel  mouse click selecting any treeview item as mouse selection is held


lukefromdc:
#### Expected behaviour
Selecting treeview rows with the mouse should work the same as it does with the arrow keys, that is clicking on a new row should move the selection to that row

#### Actual behaviour
Selecting rows with arrow keys works, but if a row is selected with the mouse the treeview does not respond to clicking on any other row (or any nother widget, or another application etc), and there is no way to cancel the selection except by hitting "Return" or another key that cannot be used as a keyboard shortcut. This will cancel the selection and free the mouse but is very non-obvious. If running mate-keybinding-properties from terminal, no warnings appear when using the arrow keys to select a row, but on mouse selection this warning pops up:

`(mate-keybinding-properties:19367): GLib-GObject-CRITICAL **: 15:18:48.658: Object class EggCellEditableEventBox doesn't implement property 'editing-canceled' from interface 'GtkCellEditable'

`
#### Steps to reproduce the behaviour
Open mate-control-center>keybindings, click on any row with the mouse. The mouse now cannot bne used to select anything else

#### MATE general version
1.21 from git master as of Nov 14, 2018

#### Package version
Locally compiled mate-control-center 1.21.2 from git master as of Nov 14, 2018

#### Linux Distribution
Debian Unstable

#### Link to downstream report of your Distribution
None as this is a local build

selectiveduplicate:
I can reproduce this issue on Ubuntu MATE 18.04.1, with ``mate-control-center`` version 1.20.2, and MATE general version 1.20.1.
It seems the mouse is captured until I press ``Esc`` or assign a new shortcut.
Browsing rows/selecting them works fine with keyboard, but if I press enter for a keybinding, the moue gets captured again until I press ``Esc`` or assign something.

lukefromdc:
Oh yeah-I had forgetted that the esc key gets you out of this.

sc0w:
This is an issue or the expected behavior?

![2018-11-15_10-25](https://user-images.githubusercontent.com/7734191/48543004-d5046500-e8c0-11e8-85e9-0f36ba5d9712.png)

And, according to the documentation:

![2018-11-15_10-29](https://user-images.githubusercontent.com/7734191/48543225-62e05000-e8c1-11e8-9632-42f0d5b4c3dc.png)

Anyway, I think it is more intuitive to select the item with `one click`, and edit the shortcut key with `double click`.

[...]

PR #391, opened 2018-11-15
keybindings: double click instead one click to edit keyboard shortcuts

sc0w:
Closes https://github.com/mate-desktop/mate-control-center/issues/390


sc0w:
yes, I confirm the warning here too.

lukefromdc:
That warning is also present in master so it should not block this PR

lukefromdc:
I cherrypicked this to 1.20 with
https://github.com/mate-desktop/mate-control-center/commit/e5c3e4f0e083481d490de00d9094cfe7197b4044
as this fixes a bug that could cause some users to restart their session, kill X, or even reboot by power cycle due to what seems to be a stuck mouse

PR #397, opened 2018-12-04
default-applications: Use the selected word processor for RTF as well

cwendling:
RTF files should be handled by the preferred word processor, just like ODF, DOC and OOXML.

### DIFF
--- a/capplets/default-applications/mate-da-capplet.c
+++ b/capplets/default-applications/mate-da-capplet.c
@@ -132,6 +132,7 @@ set_changed(GtkComboBox* combo, MateDACapplet* capplet, GList* list, gint type)
 
 			case DA_TYPE_WORD:
 				g_app_info_set_as_default_for_type(item, "application/vnd.oasis.opendocument.text", NULL);
+				g_app_info_set_as_default_for_type(item, "application/rtf", NULL);
 				g_app_info_set_as_default_for_type(item, "application/msword", NULL);
 				g_app_info_set_as_default_for_type(item, "application/vnd.openxmlformats-officedocument.wordprocessingml.document", NULL);
 				break;


lukefromdc:
Can you post a test file for this, for a before and after comparison?

alexarnaud:
Le 05/12/2018 à 08:05, lukefromdc a écrit :
 > Can you post a test file for this, for a before and after comparison?

You could download an RTF file here:
https://www.dropbox.com/s/7vsdhclcgnf02kl/test.rtf?dl=0

Best regards,
Alex.

cwendling:
@lukefromdc First off, it's likely not to change anything on your system, because for the setting to be really meaningful you'll need to have more than one word processor.  So to properly test you should have more than one of these: my use case was with 2 different LibreOffice versions, but possibly Abiword would also register as handling ODF and MSWord and so would appear there as well.

Anyway, to see it in effect:
1. Make sure you have more than one application handling ODF and DOC (and OOXML and RTF) installed.
2. Take any RTF file (like @alexarnaud's one), or create one with your favorite word processor.
3. Open Caja and right-click on your RTF file.  You'll see the top item be *"Open with [whichever app is the default]"*
4. Open *mate-default-applications-properties*
    1. in the *Office* tab, *change* the *Word Processor* to any value. This change is to trigger saving the MIME association, which only happens when changing the selected item.
    2. if the new value isn't the one you'd like to try, change it again to the one you really like.
5. Repeat step 3, and see the selected application change following the *Word Processor* setting.
    **Beware:** Caja doesn't re-create the popup menu every single time, so make sure to change the selected file between two right-clicks, otherwise you'll get a cached version of the popup.

lukefromdc:
I'm not going to be able to install abiword to test this, as it would add 40MB to my install or alternately require cleaning out a big pile of dependencies afterwards. Also I have no landline internet access and thus must watch bandwidth

raveit65:
I can confirm that application selection doesn't work without the PR.
Libreoffice is installed as default and later i installed abiword.
Now abiworld is always the default, doesn't matter which application is selected by  *mate-default-applications-properties*.

alexarnaud:
Le 07/12/2018 à 11:15, raveit65 a écrit :
> I can confirm that application selection doesn't work without the PR.
> Libreoffice is installed as default and later i installed abiword.
> Now abiworld is always the default, doesn't matter which application is 
> selected by /mate-default-applications-properties/.
> 

Do you mean that it works properly or not? I'm not sure to understand 
correctly your message.

Best regards,
Alex.

raveit65:
I consider this is a bugfix and i will cherry-pick it to 1.20 branch.

PR #398, opened 2018-12-06
default-applications: Add a widget for a default calculator application

zhangxianwei8:
• Exposes a widget in Preferred Applications to control the default calculator application schema defined in commit https://github.com/mate-desktop/mate-desktop/pull/341/commits/328116089489c674bad3b17bc69c267865968d24
• In order to function properly, this PR is dependent on the merger of these two PRs:
1. https://github.com/mate-desktop/mate-desktop/pull/341
2. https://github.com/mate-desktop/mate-settings-daemon/pull/255

• Also fix https://github.com/mate-desktop/mate-control-center/issues/291

PR #399, opened 2018-12-08
default-applications: Fix assertion failture on g_icon_to_string

zhangxianwei8:
Prevent runtime failture on g_icon_to_string in fill_combo_box by catching the NULL case

See https://github.com/mate-desktop/mate-control-center/pull/398#issuecomment-445047978

### DIFF
--- a/capplets/default-applications/mate-da-capplet.c
+++ b/capplets/default-applications/mate-da-capplet.c
@@ -501,12 +501,16 @@ fill_combo_box(GtkIconTheme* theme, GtkComboBox* combo_box, GList* app_list, gch
 		
 		/* Icon */
 		GIcon* icon = g_app_info_get_icon(item);
-		gchar* icon_name = g_icon_to_string(icon);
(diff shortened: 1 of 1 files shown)


capplets/default-applications/mate-da-capplet.c
```diff
@@ -501,12 +501,16 @@ fill_combo_box(GtkIconTheme* theme, GtkComboBox* combo_box, GList* app_list, gch
...
 		GIcon* icon = g_app_info_get_icon(item);
-		gchar* icon_name = g_icon_to_string(icon);
-		
-		if (icon_name == NULL)
-		{
-			/* Default icon */
-			icon_name = g_strdup("binary");
+		gchar* icon_name;
+
+		if (icon != NULL) {
```
muktupavels: Would not it be better to just add two newlines?
```
if (icon != NULL)
  icon_name = g_icon_to_string (icon);
```

PR #404, opened 2018-12-29
Migrate from dbus-glib to GDBus

yetist:
mate-display-properties: capplets/display/xrandr-capplet.c

Note that if you look at `mate-display-properties` executable with `ldd`, you will still find that it depends on `libdbus-glib`. Because I didn't modify Makefile.am, I thought it would be better to migrate all the other code to gdbus and then modify makefile.am.

Please review the code and test it, thank you.

### DIFF
--- a/capplets/display/xrandr-capplet.c
+++ b/capplets/display/xrandr-capplet.c
@@ -35,8 +35,6 @@
 #include <X11/Xlib.h>
 #include <glib/gi18n.h>
 #include <gio/gio.h>
-#include <dbus/dbus-glib.h>
-#include <dbus/dbus-glib-bindings.h>
 
 typedef struct App App;
 typedef struct GrabInfo GrabInfo;
@@ -71,9 +69,8 @@ struct App
     GSettings	   *settings;
 
     /* These are used while we are waiting for the ApplyConfiguration method to be executed over D-bus */
-    DBusGConnection *connection;
-    DBusGProxy *proxy;
-    DBusGProxyCall *proxy_call;
+    GDBusConnection *connection;
+    GDBusProxy *proxy;
 
     enum {
 	APPLYING_VERSION_1,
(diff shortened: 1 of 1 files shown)


lukefromdc:
It would seem to me that Makefile.am should be updated now so that the new code actually links to the library it will use. Otherwise I would expect trouble on trying to actually use the applet

yetist:
OK, I will do it.

raveit65:
Here it works fine without a change in make file.
About which make file are you talking?
I see `dbus-glib-1` only in configure.ac

yetist:
`mate-display-properties` use `capplets/common/Makefile.am`, but many other applets also use it, and they  require `dbus-glib-1` now, so I want to keep the current state until we finished migrate to gdbus.

raveit65:
Agree, i see no problem to run `mate-display-properties` with this PR.

capplets/display/xrandr-capplet.c
```diff
@@ -1938,48 +1935,61 @@ static void
...
-    g_assert (app->proxy != NULL); /* that call does not fail unless we pass bogus names */
+    app->proxy = g_dbus_proxy_new_sync (app->connection,
+                                        G_DBUS_PROXY_FLAGS_NONE,
+                                        NULL,
+                                        "org.mate.SettingsDaemon",
+                                        "/org/mate/SettingsDaemon/XRANDR",
+                                        "org.mate.SettingsDaemon.XRANDR_2",
+                                        NULL,
+                                        &error);
+    if (app->proxy == NULL) {
```
muktupavels: Why did you replace assert with this? In case you have null proxy you will do this `g_dbus_proxy_call (NULL, ...)`...

raveit65:
Tested again --> runs fine!

raveit65:
@yetist 
It might be a good idea to reword the commits to avoid similar naming with other caplets in git log?
`Migrate from dbus-glib to GDBus` --> `m-d-p:  Migrate from dbus-glib to GDBus`
(m-d-p = mate-display-properties)
Similar for second commit.

yetist:
> @yetist
> It might be a good idea to reword the commits to avoid similar naming with other caplets in git log?
> `Migrate from dbus-glib to GDBus` --> `m-d-p: Migrate from dbus-glib to GDBus`
> (m-d-p = mate-display-properties)
> Similar for second commit.

Good idea, I will do it later.

lukefromdc:
According to 
https://developer.gnome.org/gio/stable/GDBusProxy.html#g-dbus-proxy-call
a NULL call won't cause any problems: 
`
If callback is NULL then the D-Bus method call message will be sent with the G_DBUS_MESSAGE_FLAGS_NO_REPLY_EXPECTED flag set.`

muktupavels:
@lukefromdc Read my comment again, more carefully...

PR #406, opened 2018-12-29
capplets/mouse: New doubleclick artwork with mate logo

sc0w:
In the past, I opened the PR https://github.com/mate-desktop/mate-control-center/pull/331

In this new PR, the images are improved, with transparent background and better appearance

As I mentioned in older PR:

My intention with this PR is to avoid the gnome background, and the light bulb behavior is not clear for me

In the light bulb, the images _double-click-maybe.png_ and _double-click-on.png_ have the bulb on,
if double click doesn't work, the user see _double-click-maybe.png_ and the bulb is on

The mate logo is the first thing the new mate user knows, ubuntu mate have other logo, but all the ubuntu mate users see the mate desktop logo in the "appearance preferences".


lukefromdc:
Here is how it looks with my proposed text clarification for user unfamiliar with the MATE logo:

mouse_properties_w_change

raveit65:
I think it could be better scaled (more sharp) for HIDPI.
Maybe a svg is a better solution?

sc0w:
@lukefromdc @raveit65 ok, I will do your suggested changes

Issue #403, opened 2018-12-09
mate-control-center crashes if installing an alternative (second) calculator


raveit65:
#### Expected behaviour
An open m-c-c window shouldn't crash if you install another calculator.

#### Actual behaviour
An open m-c-c window crashes if you install another calculator.

#### Steps to reproduce the behaviour
1. open m-c-c from terminal
2. Install another calculator with you package-manager, eg. `dnf install gnome-calculator` for fedora
3. Boom, the open m-c-c window segfaults.
```
[rave@mother ~]$ LANG=C
[rave@mother ~]$ mate-control-center 

** (mate-control-center:2345): WARNING **: 12:13:10.131: 
error raised: [load_xbel_store: couldn't load bookmark file [NULL]
]


(mate-control-center:2345): GLib-GObject-WARNING **: 12:14:15.382: invalid unclassed pointer in cast to 'GtkWidget'

(mate-control-center:2345): Gtk-CRITICAL **: 12:14:15.382: gtk_widget_destroy: assertion 'GTK_IS_WIDGET (widget)' failed

<cut>

** (mate-control-center:2345): CRITICAL **: 12:14:15.442: matemenu_tree_get_root_directory: assertion 'tree->loaded' failed

(mate-control-center:2345): GLib-GObject-WARNING **: 12:14:15.448: invalid unclassed pointer in cast to 'GtkWidget'

(mate-control-center:2345): Gtk-CRITICAL **: 12:14:15.448: _gtk_widget_get_preferred_size_and_baseline: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-control-center:2345): GLib-GObject-WARNING **: 12:14:15.448: invalid unclassed pointer in cast to 'GtkTable'

(mate-control-center:2345): GLib-GObject-WARNING **: 12:14:15.448: invalid unclassed pointer in cast to 'GtkContainer'

(mate-control-center:2345): Gtk-CRITICAL **: 12:14:15.448: gtk_container_foreach: assertion 'GTK_IS_CONTAINER (container)' failed

(mate-control-center:2345): GLib-GObject-WARNING **: 12:14:15.448: invalid unclassed pointer in cast to 'GtkContainer'

(mate-control-center:2345): Gtk-CRITICAL **: 12:14:15.448: gtk_container_foreach: assertion 'GTK_IS_CONTAINER (container)' failed

[...]

PR #408, opened 2018-12-30
Fixed crash when install/remove application.

yetist:
Fixed issue #403

If m-c-c is open, install/uninstall application, m-c-c is crash.

### DIFF
--- a/libslab/app-resizer.c
+++ b/libslab/app-resizer.c
@@ -136,11 +136,11 @@ calculate_num_cols (AppResizer * resizer, gint avail_width)
 			GtkTable *table = GTK_TABLE (resizer->cached_tables_list->data);
 			GList *children = gtk_container_get_children (GTK_CONTAINER (table));
 			GtkWidget *table_element = GTK_WIDGET (children->data);
-			GtkAllocation allocation;
+			gint natural_width;
 			g_list_free (children);
 
-			gtk_widget_get_allocation (table_element, &allocation);
-			resizer->cached_element_width = allocation.width;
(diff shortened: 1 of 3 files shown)


raveit65:
@yetist 
You can create branches for a PR directly here in this repo now.
No need to fork the repo first  :wink:

raveit65:
I can confirm that the PR still fixes the crash, but the repeating warnings remain in terminal if i install/uninstall an application.

```
(mate-control-center:19304): GLib-GObject-WARNING **: 12:09:19.499: invalid unclassed pointer in cast to 'GtkWidget'

(mate-control-center:19304): Gtk-CRITICAL **: 12:09:19.499: gtk_widget_destroy: assertion 'GTK_IS_WIDGET (widget)' failed

<cut>
```
Good is that the warnings aren't in in xsession-errors or journal, so we don't spam logs.

raveit65:
Ok, the remaining warnings are still there with fedora 28 and m-c-c-1.20.3.
So the crash is fixed well.

yetist:
The bugs I found in #403 has been fixed now, please review the code and test it, thank you.

Note: The three commits fix different issues, so should keep that, don't squash them.

raveit65:
I tested it again all warnings are gone. Thanks a lot.

yetist:
Yes, there is a memory leak, now it fixed.

raveit65:
Works fine in another test.

yetist:
> @yetist
> Warnings are still there in 1.20 branch. I guess i need to back port only the last 2 commits ?

Yes, I think so.

PR #410, opened 2019-01-30
default-applications: Use more common icon names for Office documents

cwendling:
_"x-office-document"_ and _"x-office-spreadsheet"_ are more common MIME icon names than their counterparts without the _"x-"_ prefix.
The latter seems not to even be available in _"MATE"_ icon theme, but only in _"Faenza"_ icon theme.  The _"x-"_-prefixed variant on the other hand is available even in theme providing the unprefixed ones, meaning it will sill work with these.

On my machine with not many themes, but only standard-ish ones:
```console
$ find /usr/share/icons/ -name office-document.* | wc -l
8
$ find /usr/share/icons/ -name x-office-document.* | wc -l
54
```

Thus, using the prefixed variant seems like a better choice.

### DIFF
--- a/capplets/default-applications/mate-da-capplet.c
+++ b/capplets/default-applications/mate-da-capplet.c
@@ -335,8 +335,8 @@ static struct {
 	{"text_image",         "text-editor"},
 	{"terminal_image",     "terminal"},
 	{"document_image",     "application-pdf"},
-	{"word_image",         "office-document"},
-	{"spreadsheet_image",  "office-spreadsheet"},
+	{"word_image",         "x-office-document"},
+	{"spreadsheet_image",  "x-office-spreadsheet"},
 	{"calculator_image",   "accessories-calculator"},
 };
 


cwendling:
@yetist Thanks for the review!

There is a similar issue with `application-pdf`, but there isn't a simple solution there: some themes only provide `gnome-mime-application-pdf`, but most others only have `application-pdf`.  As the `gnome-mime-` prefix doesn't seem standard to me, I'd rather see the themes providing a link named `application-pdf` (_"mate"_ theme notably provides both).

lukefromdc:
I could see the result of the PDF issue in this PR: it fixes other missing icons in the MATE theme, but in the case of the Adwaita icon theme the icon for a PDF file is missing both with and without this PR.  Still, this fixes missing icons in the mate icon themes entirely,
m-c-c_master_vs_pr_3_icon_theme

lukefromdc:
I see the code change here does not touch pdf icon handling, so this is good to go

PR #414, opened 2019-02-02
Travis support

yetist:
I think it's ready to go, it is currently supported to compile debian sid, fedora 29 and ubuntu 18.10 at the same time.

We only focus on the compilation process, and other output information is as small as possible. `make distcheck` is run, but only on debian can't pass, so `make distcheck` is allowed to run, but don't change the return value.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -137,12 +137,14 @@ before_scripts:
   - if [ ${DISTRO_NAME} == "debian" -o ${DISTRO_NAME} == "ubuntu" ];then ./configure --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu --libexecdir=/usr/lib/x86_64-linux-gnu  >/dev/null; else ./configure --prefix=/usr  >/dev/null; fi
   - make > /dev/null
   - make install >/dev/null
-  - popd
 
   - cd ${START_DIR}
-  - "[ -f mate-menus-1.21.0.tar.xz ] || curl -o mate-menus-1.21.0.tar.xz http://pub.mate-desktop.org/releases/1.21/mate-menus-1.21.0.tar.xz"
-  - tar xf mate-menus-1.21.0.tar.xz
-  - cd mate-menus-1.21.0
+    #- "[ -f mate-menus-1.21.0.tar.xz ] || curl -o mate-menus-1.21.0.tar.xz http://pub.mate-desktop.org/releases/1.21/mate-menus-1.21.0.tar.xz"
+    #- tar xf mate-menus-1.21.0.tar.xz
+    #- cd mate-menus-1.21.0
+  - git clone --depth 1  https://github.com/mate-desktop/mate-menus.git
+  - cd mate-menus
+  - ./autogen.sh >/dev/null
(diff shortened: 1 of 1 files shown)


raveit65:
I am still wondering about the hard coded versions in `before_scripts` section.
```
+before_scripts:
+  # These scripts should keep silent.
+  - cd ${START_DIR}
+  - '[ -f mate-desktop-1.21.2.tar.xz ] || curl -Ls -o mate-desktop-1.21.2.tar.xz http://pub.mate-desktop.org/releases/1.21/mate-desktop-1.21.2.tar.xz'
+  - tar xf mate-desktop-1.21.2.tar.xz
+  - cd mate-desktop-1.21.2
+  - if [ ${DISTRO_NAME} == "debian" -o ${DISTRO_NAME} == "ubuntu" ];then ./configure --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu --libexecdir=/usr/lib/x86_64-linux-gnu  >/dev/null 2>&1; else ./configure --prefix=/usr  >/dev/null 2>&1; fi
+  - make > /dev/null 2>&1
+  - make install >/dev/null 2>&1
+
+  - cd ${START_DIR}
+  - git clone --depth 1  https://github.com/mate-desktop/mate-menus.git
+  - cd mate-menus
+  - ./autogen.sh >/dev/null 2>&1
+  - if [ ${DISTRO_NAME} == "debian" -o ${DISTRO_NAME} == "ubuntu" ];then ./configure --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu --libexecdir=/usr/lib/x86_64-linux-gnu  >/dev/null 2>&1; else ./configure --prefix=/usr  >/dev/null 2>&1; fi
+  - make > /dev/null 2>&1
+  - make install > /dev/null 2>&1
+
+  - cd ${START_DIR}
+  - '[ -f mate-settings-daemon-1.21.2.tar.xz ] || curl -Ls -o mate-settings-daemon-1.21.2.tar.xz http://pub.mate-desktop.org/releases/1.21/mate-settings-daemon-1.21.2.tar.xz'
+  - tar xf mate-settings-daemon-1.21.2.tar.xz
+  - cd mate-settings-daemon-1.21.2
+  - if [ ${DISTRO_NAME} == "debian" -o ${DISTRO_NAME} == "ubuntu" ];then ./configure --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu --libexecdir=/usr/lib/x86_64-linux-gnu  >/dev/null 2>&1; else ./configure --prefix=/usr  >/dev/null 2>&1; fi
+  - make > /dev/null 2>&1
+  - make install > /dev/null 2>&1
+
```
I bang my head and maybe i miss something, is it similar like the usage of a minimal required version?

yetist:
Yes, it requires these versions, and mate-menus-1.21 is indirectly dependent:
```
MSD_REQUIRED=1.21.2

MATEKBD_REQUIRED=1.17.0

MATE_DESKTOP_REQUIRED=1.21.2
```

raveit65:
Ok, than this extra downloads are still needed until packages which satisfy the minimum requires are in distros.

PR #423, opened 2019-03-11
control-center: Fix memory leak

rbuj:
Reported by clang:

  control-center.c:84:2: warning: Potential leak of memory pointed to by 'action'
          g_slist_free(key_list);
          ^~~~~~~~~~~~

### DIFF
--- a/shell/control-center.c
+++ b/shell/control-center.c
@@ -70,6 +70,8 @@ static GSList* get_actions_list(void)
 
 		if ((action->item = load_desktop_item_from_unknown(temp[1])) == NULL)
 		{
+			g_free (action->name);
+			g_free (action);
 			g_warning("get_actions_list() - PROBLEM - Can't load %s\n", temp[1]);
 		}
 		else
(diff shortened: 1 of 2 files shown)


shell/control-center.c
```diff
@@ -70,6 +70,8 @@ static GSList* get_actions_list(void)
 
 		if ((action->item = load_desktop_item_from_unknown(temp[1])) == NULL)
 		{
+			g_free ((char*)action->name);
```
muktupavels: Perhaps `name` should not be const in first place?

rbuj: thanks for the tip

shell/control-center.c
```diff
@@ -81,7 +83,7 @@ static GSList* get_actions_list(void)
 		g_free(entry);
 	}
 
-	g_slist_free(key_list);
+	g_slist_free (key_list);
```
muktupavels: Looks like unrelated change...

rbuj: I appreciate your comments

PR #430, opened 2019-03-21
default-applications: Improve UI accessibility

cwendling:
Replace plain labels with frames as those properly group and label
their contents for assistive technologies.

There are no visual changes.

### DIFF
--- a/capplets/default-applications/mate-default-applications-properties.ui
+++ b/capplets/default-applications/mate-default-applications-properties.ui
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.0 -->
+<!-- Generated with glade 3.22.1 -->
 <interface>
   <requires lib="gtk+" version="3.14"/>
   <!-- interface-license-type gplv2 -->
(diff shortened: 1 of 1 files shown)


raveit65:
Looks like before with all our themes and HighContrast themes, so i don't get what your plan is with this commit.
Any way, seems OK.

cwendling:
> Looks like before with all our themes and HighContrast themes, so i don't get what your plan is with this commit.

The goal was to get a semantic structure so assistive technologies like a screen reader can tell that for example "Web Browser" labels the combo box beneath it.  Without this, it's just a list of widgets with no relations between them, and thus a screen reader can't tell what's the purpose of which combo box.
With this change, "Web Browser" becomes the label of the container that has the web browser combo box as one of its children, and thus the screen reader can tell what's what.

Basically, before the change, the widget tree looked like this:
```
Internet Notebook Page
|- Label ("Web Browser")
|- Box
|  |- Image
|  \- Combo Box
|- Label ("Mail Reader")
\- Box
   |- Image
   \- Combo Box
```

Now:
```
Internet Notebook Page
|- Frame ("Web Browser")
|  \- Box
|     |- Image
|     \- Combo Box
\- Frame ("Mail Reader")
   \- Box
      |- Image
      \- Combo Box
```

As you see, before the change labels and boxes are at the same level.  Now, boxes are children of frames, which are properly labeled.

This has the advantage over simply creating a relation between the label and the combo box that this method gives *context* to the children widgets, and so e.g. a screen reader can avoid repeating the label for each children, yet knows the relation for all of them so can speak it when appropriate (e.g. when context changes).

PR #444, opened 2019-05-12
Fix the icon sizes in Preferred Applications selector

### LINKED ISSUE #358: Default Applications shows icons at different sizes, causing misalignment
#### Actual behaviour

sil-1009810219822665728-dgorbd1xuael8-s

#### Expected behaviour

The dropdown lists should be the same width.

Open "Preferred Applications".

Whether this problem exhibits is probably dependent on icon themes, but the basic issue is that the icons should be displayed at the same width, regardless. It would be fine I think to display a smaller icon inside an invisible box the width of a larger icon, but the key point is that the dropdown lists not lining up looks ugly.

I think this is fixable in `mate-control-center/master/capplets/default-applications/mate-default-applications-properties.ui` but I'm not sure how best to tell a Gtk Image to always be a certain explicit width and to display the "actual" image centred in that width, or whether there's a better way to solve it.

#### MATE general version

1.20.1.

#### Package version

1.20.2-2ubuntu1

#### Linux Distribution

Ubuntu 18.04.

tamplan:
Fixes #358

### DIFF
--- a/capplets/default-applications/mate-da-capplet.c
+++ b/capplets/default-applications/mate-da-capplet.c
@@ -302,7 +302,8 @@ refresh_combo_box_icons(GtkIconTheme* theme, GtkComboBox* combo_box, GList* app_
 				          ICONAME_COL, &icon_name,
 				          -1);
 
-		pixbuf = gtk_icon_theme_load_icon(theme, icon_name, 22, 0, NULL);
+		GtkIconInfo* icon_info = gtk_icon_theme_lookup_icon (theme, icon_name, 22, GTK_ICON_LOOKUP_FORCE_SIZE);
+		pixbuf = gtk_icon_info_load_icon(icon_info, NULL);
 
 		gtk_list_store_set(GTK_LIST_STORE(model), &iter,
 				PIXBUF_COL, pixbuf,
@@ -330,7 +331,7 @@ static struct {
 	{"mobility_image",     "preferences-desktop-accessibility"},
 	{"messenger_image",    "user-idle"},
 	{"filemanager_image",  "file-manager"},
-	{"imageviewer_image",  "image-x-generic"},
+	{"imageviewer_image",  "eom"},
 	{"video_image",        "video-x-generic"},
 	{"text_image",         "text-editor"},
 	{"terminal_image",     "terminal"},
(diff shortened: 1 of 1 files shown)


raveit65:
This looks better:)
I will test it.

raveit65:
Icons are looking fine, i am only wondering why i can't select eom anymore as image-viewer?

flexiondotorg:
I don't experience the missing icon for Firefox in Ubuntu MATE, only image-viewer isn't being found.

raveit65:
Ok, we need to update desktop file of eom.
Editing `MimeType` entry helps a lot.
```
-MimeType=image/*;
+MimeType=image/bmp;image/jpeg;image/gif;image/png;image/tiff;image/x-bmp;image/x-ico;image/x-png;image/svg+xml;
```
and updating desktop database
`update-desktop-database -v /usr/share/applications/`
makes eom selectable in mate-default-applications-properties.

Note: this is an example. We need to check which `MimeTypes` eom supports.

Firefox is selectable here.

raveit65:
> only image-viewer isn't being found.

Gthumb or Gimp is being find out of box here.

capplets/default-applications/mate-da-capplet.c
```diff
@@ -330,7 +331,7 @@ static struct {
 	{"mobility_image",     "preferences-desktop-accessibility"},
 	{"messenger_image",    "user-idle"},
 	{"filemanager_image",  "file-manager"},
-	{"imageviewer_image",  "image-x-generic"},
+	{"imageviewer_image",  "eom"},
```
monsta: Was it needed to change this icon name? Most icon themes out there don't have a `eom` icon...

tamplan: Hi,
Because Eye Of Mate is part of Mate Desktop, nothing else.

Regards

lukefromdc:
In fact,  EOM has been one of the problem apps that has often
disappeared not only as the chosen default but from the right-click
menu im caja as well. Sometimea my first sign of this has been an image
opening in GIMP by surprise when clicked on.

capplets/default-applications/mate-da-capplet.c
rbuj: The icon-name should be reverted to "image-x-generic", like "video-x-generic" and "audio-x-generic". Those icons are the generic mime types which can be opened using its default application, the user sets that in the tab below.

Captura de pantalla a 2019-06-07 23-54-41

eom isn't displayed in the list of image viewers like eog does. It's an eom issue.

tamplan: Hi,

Atril is the default office file viewer and purposed like this because it's part of the Mate Desktop, his icon is used in the GUI. In previous version of the PR, i purpose to use another generic icon (not the correct icon sure) and keep the "image-x-generic" too.

I think EOM should be presented like Atril so i purpose to use his icon too for the GUI.

Regards

rbuj: See https://github.com/mate-desktop/mate-control-center/pull/451

tamplan: Ok, seen.

PR #447, opened 2019-05-22
libslab: avoid deprecated g_type_class_add_private

yetist:


raveit65:
I am a little bit afraid to test this when i get a segfault. Any progress?

raveit65:
Indeed a segfault.
```
[root@mother rave]# mate-control-center 
Segmentation fault (core dumped)
```

yetist:
`bookmark-agent.c` is very different, so keep it use the origin code.

raveit65:
Segfault is gone. I will squash both commits.

PR #451, opened 2019-06-08
Revert icon-name for imageviewer_image

rbuj:
The icon-name should be "image-x-generic" (imageviewer_image),
like "audio-x-generic" (media_player_image) and "video-x-generic"
(video_image).

Those icons (imageviewer_image, media_player_image,
video_image) are the generic mime types which can be opened
using its preferred application.

Before:

Captura de pantalla a 2019-06-07 23-54-41

After:

Captura de pantalla a 2019-06-08 09-24-36


monsta:
Thanks, I've picked + squashed both commits (#444 and this) to 1.22 and 1.20 branches.

PR #468, opened 2019-06-18
time-admin: shortcuts, images & shared code for buttons

rbuj:
- Add & enable mnemonic accelerator keys
- Show Save/Close button icons in Time/Date manager window
- Remove DialogAddButtonWithIconName from shares

### DIFF
--- a/capplets/time-admin/src/time-zone.c
+++ b/capplets/time-admin/src/time-zone.c
@@ -426,9 +426,7 @@ static void ChoooseTimezoneClose(GtkWidget  *widget,
 
 void SetupTimezoneDialog(TimeAdmin *ta)
 {
-    GtkWidget *Vbox;
-    GtkWidget *TimeZoneFrame;
-    GtkWidget *Scrolled;
+    GtkWidget *Vbox, *TimeZoneFrame, *Scrolled, *image;
 
     ta->dialog = gtk_dialog_new_with_buttons (_("Time Zone Selection"),
                                               NULL,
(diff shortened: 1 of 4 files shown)


capplets/time-admin/src/time-zone.c
```diff
@@ -437,15 +437,21 @@ void SetupTimezoneDialog(TimeAdmin *ta)
...
-                                               GTK_RESPONSE_OK);
+
+    ta->TZclose = gtk_button_new_with_mnemonic (_("_Close"));
+    gtk_button_set_image (GTK_BUTTON (ta->TZclose), gtk_image_new_from_icon_name ("window-close", GTK_ICON_SIZE_BUTTON));
+    gtk_button_set_use_underline (GTK_BUTTON (ta->TZclose), TRUE);
+    gtk_style_context_add_class (gtk_widget_get_style_context (ta->TZclose), "text-button");
+    gtk_widget_set_can_default (ta->TZclose, TRUE);
+    gtk_dialog_add_action_widget (GTK_DIALOG (ta->dialog), ta->TZclose, GTK_RESPONSE_CANCEL);
+
+    ta->TZconfire = gtk_button_new_with_mnemonic (_("Con_firm"));
```
zhuyaliang: It should be ("_Confirm")

rbuj: You can't use the same the same key accelerator twice.

zhuyaliang:
Why not encapsulate it as a function? Is there any benefit in doing so?

rbuj:
You are right, we can see two blocks of code that they seem to be duplicated, as we build two buttons with similar code. It's done on form build, and the code encapsulation here can obfuscate the initialization of the controls. Do you want a private function/method to build the buttons?

zhuyaliang:
This ```DialogAddButtonWithIconName``` function can fully meet the needs, why remove it?Encapsulated in the form of functions, the code has better readability.

rbuj:
> This `DialogAddButtonWithIconName` function can fully meet the needs, why remove it?Encapsulated in the form of functions, the code has better readability.

In the same manner you didn't encapsulate buttons build in `static GtkWidget *SetDate(TimeAdmin *ta);`, and`DialogAddButtonWithIconName` should be local.

zhuyaliang:
Thank you for adding icons to the buttons, but what I want to say is that a lot of duplicated code is not beautiful, and I think they can be encapsulated as a common function.

rbuj:
Yes, the next natural step could be to write the ui files using glade.

rbuj:
WIP time-admin.ui - Only for testing:
https://github.com/mate-desktop/mate-control-center/tree/test-time-admin-ui

PR #472, opened 2019-06-22
[ci] add gla11y in docker container

yetist:
Let us see what happen.


sthibaul:
It seems to be going well, except a missing cleanup, I have submitted https://github.com/mate-desktop/mate-control-center/pull/473 for that

sthibaul:
I'd say that, like compiler warnings and errors, we'd better go with taking a tagged snapshot, like
https://github.com/hypra/gla11y/blob/v0.2/gla11y
so that if gla11y introduces new fatals in its master branch, the mate-control-center CI won't suddenly break.

sthibaul:
Also, to have better line number reporting, it'd be useful to add the python-lxml dependency.

yetist:
I think it's ready to go now.

sthibaul:
It looks good indeed. Perhaps you could even remove the existing `$(GLA11Y_OUTPUT)` from CLEAN, as the attached patch does: 
[patch.txt](https://github.com/mate-desktop/mate-control-center/files/3318581/patch.txt)

yetist:
@sthibaul 

Done, thanks. 5781cdb

yetist:
Squash the commits into one.

PR #473, opened 2019-06-22
gla11y: Fix missing output cleanup

sthibaul:


yetist:
Is that possible to add the clean target into the gla11y.mk for all files?

yetist:
> Is that possible to add the clean target into the gla11y.mk for all files?

I did is by this commit: https://github.com/mate-desktop/mate-control-center/pull/472/commits/3c99b8081dce710dec2d1de9956f2b30c09f20ad , and it seems woks.

sthibaul:
This will actually be replaced by https://github.com/mate-desktop/mate-control-center/pull/472 's support

PR #476, opened 2019-06-23
libslab: bookmark-agent.c avoid deprecated g_type_class_add_private

yetist:
Please review and test, thanks.


raveit65:
This runs fine but i am not sure if we really use the bookmark agent because of this old warning (exists with master).
```
[rave@mother ~]$ mate-control-center 

** (mate-control-center:4769): WARNING **: 13:40:21.334: 
error raised: [load_xbel_store: couldn't load bookmark file [NULL]
]
```

PR #479, opened 2019-06-27
at-properties: migrate from dbus-glib to gdbus

yetist:
Please review and test.

### DIFF
--- a/capplets/accessibility/at-properties/main.c
+++ b/capplets/accessibility/at-properties/main.c
@@ -108,64 +108,59 @@ cb_login_preferences (GtkDialog *dialog, gint response_id)
     }
 }
 
-/* get_session_bus(), get_sm_proxy(), and do_logout() are all
- * based on code from mate-session-save.c from mate-session.
- */
-static DBusGConnection *
-get_session_bus (void)
-{
-        DBusGConnection *bus;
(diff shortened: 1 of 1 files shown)


raveit65:
@cwendling 
Does this work for you?

capplets/accessibility/at-properties/main.c
```diff
@@ -86,64 +86,59 @@ cb_login_preferences (GtkDialog *dialog, gint response_id)
...
+
+    ret = g_dbus_proxy_call_sync (sm_proxy,
+                                  "Logout",
+                                  g_variant_new ("(u)", 0),    /* '0' means 'log out normally' */
+                                  G_DBUS_CALL_FLAGS_NONE,
+                                  -1,
+                                  NULL,
+                                  &error);
+    if (ret == NULL) {
+        g_propagate_error (err, error);
```
vkareh: This error is propagated here, but `do_logout` is only ever called with `NULL`. I would suggest changing the signature of `do_logout` to remove `err` and change this block to:
```
g_warning ("...", error->message);
g_error_free (error);
return FALSE;
```
(or something along those lines)

raveit65:
Only a rebase from master after latest merge.

PR #482, opened 2019-07-08
Use python3 for gla11y

sthibaul:


sthibaul:
(since python3 is being phased out)

Issue #481, opened 2019-07-08
libXxf86misc dependency becomes deprecated


raveit65:
#### Expected behaviour
Building package with libXxf86misc dependency in fedora 31 (next release in October /November)

#### Actual behaviour
libXxf86misc dependency can't be satisfied in fedora 31

#### Steps to reproduce the behaviour
Complie package in fedora rawhide (f31)

#### MATE general version
All

#### Package version
All

#### Linux Distribution
Fedora, probably other will follow bleeding edge leader

#### Link to downstream report of your Distribution
Not reported, things needs to be clarified before i file out a report against libXxf86misc.
Latest version of libXxf86misc in fedora.
https://koji.fedoraproject.org/koji/buildinfo?buildID=1291079

Comment from Xorg developer:
```
Stop building the devel subpackage in Fedora and RHEL > 8,
no new code should try to use this library,
the X server hasn't implemented it in over 10 years
```
In result i have to build mate-settings-demon, mate-screensaver and mate-control-center without it.
I have no idea about the consequences.
Does anyone know for what libXxf86misc is used?
Or better for what we use it in mate-control-center?

@mate-desktop/contributors @mate-desktop/core-team @mate-desktop/documentation-team 
Any idea?

lukefromdc:
I can't even find this package on my system, the only packages I find in Debian Unstable with names containing "libXxf86" are the two below:

libxxf86dga 
"libXxf86dga provides the XFree86-DGA extension, which allows direct
graphics access to a framebuffer-like region, and also allows relative
mouse reporting, et al.  It is mainly used by games and emulators for
games."

libXxf86vm 
libXxf86vm provides an interface to the XFree86-VidModeExtension extension, which allows client applications to get and set video mode
timings in extensive detail.  It is used by the xvidtune program in particular."

raveit65:
I think it is removed from debian since a long time https://github.com/mate-desktop/mate-dev-scripts/issues/23
So, it is save to remove the dependency from configure.ac.

raveit65:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=568285

PR #485, opened 2019-07-10
drop libXxf86 dependency

raveit65:
The X server hasn't implemented it in over 10 years.
and it was dropped from debian since a long time.

fixes https://github.com/mate-desktop/mate-control-center/issues/481


raveit65:
Thanks for the review :)

PR #487, opened 2019-07-15
[appearance] Ident by 4 spaces and code refactoring

tamplan:
Change GtkWidget to GtkDialog for dialog
Add brackets for all for, while, if statements including one line blocks code
Better use of gboolean in one function


tamplan:
Change two switch blocks.

PR #488, opened 2019-07-15
[appearance] Ident by 4 spaces and code refactoring

tamplan:
Change GtkWidget to GtkDialog for dialog several times.
Add brackets for all for, while, if statements including one line blocks code.
Better use of gboolean.
Change two switch blocks.

### DIFF
--- a/capplets/appearance/appearance-desktop.c
+++ b/capplets/appearance/appearance-desktop.c
@@ -31,424 +31,459 @@
 #include <libmate-desktop/mate-bg.h>
 
 enum {
-	TARGET_URI_LIST,
-	TARGET_BGIMAGE
+    TARGET_URI_LIST,
+    TARGET_BGIMAGE
 };
 
 static const GtkTargetEntry drop_types[] = {
-	{"text/uri-list", 0, TARGET_URI_LIST},
-	{"property/bgimage", 0, TARGET_BGIMAGE}
+    {"text/uri-list", 0, TARGET_URI_LIST},
+    {"property/bgimage", 0, TARGET_BGIMAGE}
 };
 
 static const GtkTargetEntry drag_types[] = {
-	{"text/uri-list", GTK_TARGET_OTHER_WIDGET, TARGET_URI_LIST}
+    {"text/uri-list", GTK_TARGET_OTHER_WIDGET, TARGET_URI_LIST}
 };
 
 
-static void wp_update_preview(GtkFileChooser* chooser, AppearanceData* data);
+static void wp_update_preview (GtkFileChooser*  chooser,
+                               AppearanceData*  data);
 
-static void select_item(AppearanceData* data, MateWPItem* item, gboolean scroll)
+static void
+select_item (AppearanceData*    data,
+             MateWPItem*        item,
+             gboolean           scroll)
 {
-	GtkTreePath* path;
+    GtkTreePath* path;
 
-	g_return_if_fail(data != NULL);
+    g_return_if_fail(data != NULL);
 
-	if (item == NULL)
-		return;
-
-	path = gtk_tree_row_reference_get_path(item->rowref);
-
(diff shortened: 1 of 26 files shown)


vkareh:
I'm not a fan of this... you're making functional changes ("Change GtkWidget to GtkDialog") along with sweeping code styling changes ("Ident by 4 spaces") in the same commit. This makes it impossible to do a proper code review.

Also it messes with the git annotations, when working on the code, I regularly use `git blame` to understand motivation behind specific lines of code. Whenever the hash points to a code styling change it means I need to manually dig through the git history to make sense of it.

I would prefer that code style changes are done only on code that is being functionally changed. The rest should stay as it was.

tamplan:
I was not sure of that PR, it for that reason i made it as draft.
The main problem i have is the mix between tabs and spaces for indentation in several files before i can make changes.

If i understand you correctly, you prefer:
- a PR only for code indentation
- follow maybe by a PR for code change

Can i made a PR only for indentation and add brackets for **for**, **while**, **do** and **if** for example?
Then make changes in code later?

vkareh:
That would be helpful, splitting the commits, but the problem will persist that when doing `git blame`, most lines in those files will point to that commit, which makes it harder to understand the motivation behind code change.

Regarding indentation in your editor, some of our files (not on all repos, unfortunately) have modlines
`/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*- */`
or
`* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*- */`
that should tell your editor how to format the code display so that it looks good. Sometimes you'll need to play with your editor settings, I've found... But as a team we haven't yet settled on what our code style should be. Maybe adding modelines instead of changing all indentation would alleviate most issues devs have with MATE code style?

@mate-desktop/core-team - thoughts?

tamplan:
Before purposing this PR, i have read this page https://wiki.mate-desktop.org/code_style
raveit65 spent a lot of time reviewing previous PR, indicated me "tamplan => ident!" :)

I try with geany (the editor i use) if it can add the line `* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*- */` you mention according the style i use, i don't know what to enter here so i remove these lines on each files.

raveit65:
Honestly, who should review that?
You are using a tool and reviewers should review all that by hand. No way!
I have better things to do instead of reading code lines whole the day....

I suggest to use only one file, or one *.c + *.h file in one commit.
And maximum 7 or 8 commits in one PR, you are new here!
That means all this will end up in several PRs.

Keep in mind that we have to do use `git bisect` sometimes, when a user report a regression.
When this PR is the cause of the regression we have a big problem to find the culprit code line.
Bisecting code is normal business in other projects like gnome or gtk+ too.
And we did it here sometimes.
I am still wondering that a experience developer (in other coding languages), like you said, isn't aware about this problems which can happen?

And my personal opinion is not to use tabs in new code or in a complete rewrite ;)

vkareh:
https://wiki.mate-desktop.org/code_style indicates 4 spaces for shell scripts. For the rest of the code it says:
> Use tabs after each line break, and space to align code hard to read

For example:
```
if (declaration_is_true (some_value))
{
	static char             b; // <tab> <spaces>
	int                     a; // <tab> <spaces>
	unsigned int            some_long_value; // <tab> <spaces>
	struct my_custom_struct some_misterious_struct;
}
```
The indentation is actually a \<Tab\>, whereas the alignment between the type and var name are \<Spaces\>. This ensures that the code will _always_ look good regardless if what editor you use and what configuration you have on it.

tamplan:
Ok, i will prepare a new PR...

Thanks.

vkareh:
There's still the issue of git annotations becoming annoying to use with a sweeping code style change... I don't have a solution for this :/

PR #492, opened 2019-07-29
appearance: add new font for user

yetist:
Users can add new fonts for themselves.

### DIFF
--- a/capplets/appearance/appearance-font.c
+++ b/capplets/appearance/appearance-font.c
@@ -28,6 +28,7 @@
 #include <glib/gi18n.h>
 #include <gio/gio.h>
 #include <gdk/gdkx.h>
+#include <glib/gstdio.h>
 
 #include "capplet-util.h"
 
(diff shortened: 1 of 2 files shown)


lukefromdc:
How should I test this? all fonts I have are system fonts

lukefromdc:
Unfortunately, that's Google, for whom I have a profound distrust and whose products I avoid. 

We have two approving reviews, but looking at the code I have one question: why is fontpath not freed after it is used to install the font? I don't see any later uses of it after the ` if (!g_file_copy` block

lukefromdc:
While I'm still waiting to hear back about why fontpath isn't freed, this builds fine and the dialog to add a font opens fine, though I didn't have anything to test with that does not come from Google.

raveit65:
Oh, sorry, i didn't saw the question about `fontpath isn't freed`
So i restored the branch when this is a problem.

lukefromdc:
Just built and installed from today's master with this merged, and yes, installing a new font works fine. I assume the kernel would clean up any memory leak from fontpath when the control center is closed? Still, if that is leaked we can fix it in another PR

raveit65:
@yetist 
> We have two approving reviews, but looking at the code I have one question: why is fontpath not freed after it is used to install the font? I don't see any later uses of it after the if (!g_file_copy block

Can this be a problem?

lukefromdc:
Could be a memory leak. While nobody is likely to keave the appearance
capplet open a week and.run this over and over, memory leaks.are still bad
practice. If.nothing else is.freeimg this we can handle it in anither PR

yetist:
@raveit65 

Yes, this has a memory leak and I will create a new PR to fix it.

lukefromdc:
https://github.com/mate-desktop/mate-control-center/pull/493 works, just needs a second review

PR #494, opened 2019-08-02
mouse: Add support for Acceleration Profiles

### LINKED ISSUE #491: [Request] Add additional mouse settings for disabling mouse acceleration and middle click emulation [$350]
As was done in Gnome 3.22 (see https://www.reddit.com/r/linux_gaming/comments/553v1p/gnome_322_includes_option_to_disable_mouse/) we should be able to disable these settings from the GUI and not be forced to use xinput --list and xinput --set-prop.

It really is not good for gaming to have mouse acceleration and I cannot get X to respect any of my .conf files to this (even after following the "man xorg.conf.d").

Image here shows how Gnome has it in the GUI: https://imgur.com/a/DYdUO.

I am currently working around this with a self-made script that I run prior to gaming, but I feel this should be more easier to work around.

Keep up the great work with MATE and I hope we can make it better!

<bountysource-plugin>

---
There is a **[$350 open bounty](https://www.bountysource.com/issues/47571327-request-add-additional-mouse-settings-for-disabling-mouse-acceleration-and-middle-click-emulation?utm_campaign=plugin&utm_content=tracker%2F517519&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F517519&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vkareh:
Currently libinput supports Adaptive and Flat acceleration profiles. We
can use these to change mouse behavior. Synaptic touchpads are not
supported through libinput, so I have not included a UI change for this
(even though it's technically supported in gsettings).

There does not seem to be a defined behavior for a Default acceleration
profile, so it's not included.

Multi-part patch:
* https://github.com/mate-desktop/mate-desktop/pull/407
* https://github.com/mate-desktop/mate-settings-daemon/pull/293
* https://github.com/mate-desktop/mate-control-center/pull/494

Fixes #491


lukefromdc:
Once this is merged we should probably bump version here, as recommended for the other two as they all go together.

raveit65:
> Once this is merged we should probably bump version here, as recommended for the other two as they all go together.

IMHMO, here we should only bump m-s-d and mate-desktop required version

As per conversation at mate-dev channel, org.mate.peripherals-mouse.gschema.xml should be moved to m-s-d.
Than we need only to bump the version in m-s-d and m-s-d required version here.

monsta:
I'm yet to check which Synaptics properties, if any, might be corresponding to these ones. Quick test in an old VM showed me a few "Accel" ones:
```
	Device Accel Profile (244):	0
	Device Accel Constant Deceleration (245):	1.000000
	Device Accel Adaptive Deceleration (246):	1.000000
	Device Accel Velocity Scaling (247):	10.000000
```

vkareh:
@monsta that's how I started initially, just a toggle. However, if we want to support synaptics/legacy driver (whatever that is in m-s-d), I think there's a difference between Default and Adaptive. So for now I'd suggest keeping it like this for future-proofing.

monsta:
I'm just worrying about possible future renames which will cause another version bump...

vkareh:
@raveit65, @monsta - I've added the Default acceleration profile to the dropdown to match the changes in mate-desktop and m-s-d.

monsta:
Ok, for legacy driver (evdev for mouse, synaptics for touchpad) it's `Device Accel Profile` property that sets the acceleration profile. According to [docs](https://www.x.org/wiki/Development/Documentation/PointerAcceleration), there's a lot of values to choose from:

0:    classic (the default)
* similar to old behaviour, but more predictable. Selects between 'polynomial' and 'simple' based on threshold =/!= 0.

-1: none

  * no velocity-dependent pointer acceleration or deceleration. If constant deceleration is also unused, motion processing is suppressed, saving some cycles. 

1: device-dependent

  * available if the hardware driver installs it. May be coming for synaptics. 

2: polynomial

  * Scales polynomial: velocity serves as the coefficient, acceleration being the exponent. Very useable, the recommended profile. 

3: smooth linear

  * scales mostly linear, but with a smooth (non-linear) start. 

4: simple

  * Transitions between accelerated/unaccelerated, but with a smooth transition range. This has the fundamental problem of accelerating on two niveaus, on which acceleration stays independent of velocity. Traditionally the default however. 

5: power

  * accelerates by a power function. velocity is the exponent here. Adheres to threshold. Will easily get hard to control, so it is important you have properly tuned your velocity estimation. 

6: linear

  * just linear to velocity and acceleration. Simple and clean. 

7: limited

  * smoothly ascends to acceleration, maxing out at threshold, where it becomes flat (is limited).

vkareh:
That's a great find, of I'd like to add support for evdev/synaptics.

That list is too long, though, I would probably want to pare it down to Classic/Default and None, mapping Classic to Adaptive and Default, and None to Flat. If users want a different curve, that should be handled by a different tool, probably... Thoughts?

raveit65:
Honestly, i prefer 2 options, like the state of PR.
Do you really see a lot of users who like to have more options?
More options = more options that something goes wrong

PS: all this obsolete synaptics code isn't use by major linux distros. Which distro use that?
We (and linux world) are going strait and slowly to wayland, why adding more obsolete code?

vkareh:
@raveit65 brings up a good point. If you're into gaming to the point that you need configurable acceleration profiles, I doubt that you're using clunky hardware with outdated drivers... The only thing people are asking for, really, is to _disable_ acceleration, not have a KDE-long list of options. New bugs and feature requests can be opened for those if users care

raveit65:
Ok, we need t make a new m-s-d release to make travis build working.

PR #496, opened 2019-08-15
default-applications: Remove messenger options from ui file

rbuj:


lukefromdc:
Didn't we just recently add this? It is simply blank and no problem on a system like
mine with no messaging applications installed.

vkareh:
@rbuj - this PR works, but why are we removing it?

lukefromdc:
I propose that we close this and not merge it

vkareh:
Also, is there a way of de-selecting whichever messenger? In my system, I noticed that Hexchat is selected and there's no way of de-selecting it. Also, the command input is always disabled, there's no way to enter data in it for me...

rbuj:
@vkareh - I saw that we don't currently use these UI elements, and launch parameters are usually set on application preferences.
```shell
git grep messenger_command_entry
git grep messenger_terminal_checkbutton
```

raveit65:
Ok, it's only in UI files.
```
[rave@mother mate-control-center]$ grep -nr messenger_command_entry *
capplets/default-applications/mate-default-applications-properties.ui:270: <property name="mnemonic_widget">messenger_command_entry</property>
capplets/default-applications/mate-default-applications-properties.ui:278: <object class="GtkEntry" id="messenger_command_entry">

[rave@mother mate-control-center]$ grep -nr messenger_terminal_checkbutton *
capplets/default-applications/mate-default-applications-properties.ui:291: <object class="GtkCheckButton" id="messenger_terminal_checkbutton">
```
I don't think that we should show something in UI which can't be work.
But why not simply hiding this elements?
Maybe someone wants to write code for it.
```
<property name="visible">False</property>
```


@vkareh 
> Also, is there a way of de-selecting whichever messenger?

I guess this is only possible if more than one is installed and you can select another.

monsta:
> In my system, I noticed that Hexchat is selected and there's no way of de-selecting it.

Isn't it the same for other default applications comboboxes? It's not possible to set it blank once some app is chosen.

raveit65:
@lukefromdc 
Is this OK for you to remove checkbox an entry from UI file?

lukefromdc:
Fine by me, this is not something I use. My comcern was completely removing
a whole section we had just added. How exactly we configure I am not so
worried about, so long as what we ship in 1.24 is clean, functional code, and
don't run ourselves ragged adding and testing new features, than taking them
right back out again.

PR #500, opened 2019-08-20
time-admin: Don't share «void QuitApp(TimeAdmin *ta)» method

rbuj:


vkareh:
@lukefromdc the time-share seems to be a library of shared methods between the different components of the time-admin capplet.

PR #517, opened 2019-12-03
keybindings: Fix schema creation in makefile

gm10:
This fixes the contents of the first schema file getting copied into the second.

This is a similar issue as I just fixed in https://github.com/mate-desktop/mate-control-center/commit/2aa9de5d86ceb7e21cecfe79d4c85626f2538c98 - might be worth for you guys to review all the makefiles touched in the gettext transition.

### DIFF
--- a/capplets/keybindings/Makefile.am
+++ b/capplets/keybindings/Makefile.am
@@ -29,7 +29,7 @@ xmldir = $(pkgdatadir)/keybindings
 xml_in_files = 00-multimedia-key.xml.in 01-desktop-key.xml.in
 xml_DATA = $(xml_in_files:.xml.in=.xml)
 
-$(xml_DATA): $(xml_in_files)
+%.xml: %.xml.in
 	$(AM_V_GEN) GETTEXTDATADIR=$(top_srcdir) $(MSGFMT) --xml --template $< -d $(top_srcdir)/po -o $@
 pkgconfigdir = $(libdir)/pkgconfig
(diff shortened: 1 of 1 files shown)


raveit65:
> might be worth for you guys to review all the makefiles touched in the gettext transition.

Shit happens....

raveit65:
Which desktop file has wrong content in capplets/keybindings ?

gm10:
Things happen indeed, to all of us, this was not meant as a complaint. ;)

This isn't about desktop files but the keybinding schemas:
https://github.com/mate-desktop/mate-control-center/blob/8ee4c484fa87e9ceef78eeeb9c607cb59f5ad5eb/capplets/keybindings/Makefile.am#L29

raveit65:
Omg, application can crash if a gsettings key doesn't exists.
Are there more places in mate-control-center where the makefile is wrong and do you plan further PRs?
If yes, do them all in one PR please.

Thanks a lot.

gm10:
@raveit65
> Are there more places in mate-control-center where the makefile is wrong and do you plan further PRs?
> If yes, do them all in one PR please.

Nothing planned, but I'm not doing a full code review, either (hence my suggestion above), I was just fixing them as I ran into them. 

That said, I did run a quick `grep "in_files =" -R` on this repository  and the remaining makefiles seem ok. My suggestion was more about other packages where I suspect you have have been or may be planning to do the same thing. I'm not aware of any other issues of this kind at this time though.

PR #521, opened 2019-12-22
at-properties: Fix UI validation

rbuj:


raveit65:
same as with https://github.com/mate-desktop/mate-control-center/pull/525#issuecomment-568783522
application/x-gtk-builder (GTK+ Builder) https://reposcope.com/mimetype/application/x-gtk-builder
--> application/x-designer (Qt Designer file) https://reposcope.com/mimetype/application/x-designer
Do we really want to change that?
What are the consequences?

PR #528, opened 2019-12-26
Look for system themes in system data dirs

romildo:
Currently `mate-control-center` does not use the mechanisms described in the [XDG Base Directory Specification](http://www.freedesktop.org/Standards/basedir-spec) to locate available themes. In this case the list of directories to look for system themes should be retrieved from the `XDG_DATA_DIRS` environment variable. But it is still using the deprecated function [`gtk_rc_get_theme_dir`](https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/deprecated/gtkrc.c#L805) (which uses the environment variable `GTK_DATA_PREFIX`).

This affects particularly the [NixOS](https://nixos.org/) package. See https://github.com/NixOS/nixpkgs/issues/75445.

This PR changes mate control center to look for system themes in the system data dirs (defined by the `XDG_DATA_DIRS` environment variable) and is inspired in gnome-control-center.

### DIFF
--- a/capplets/common/gtkrc-utils.c
+++ b/capplets/common/gtkrc-utils.c
@@ -60,15 +60,19 @@ gchar* gtkrc_find_named(const gchar* name)
 
 	if (!path)
 	{
-		gchar* theme_dir = gtk_rc_get_theme_dir();
-		path = g_build_filename(theme_dir, name, subpath, NULL);
-		g_free(theme_dir);
+		const gchar * const * dirs = g_get_system_data_dirs();
 
-		if (!g_file_test(path, G_FILE_TEST_EXISTS))
-		{
-			g_free (path);
-			path = NULL;
-		}
+		if (dirs != NULL)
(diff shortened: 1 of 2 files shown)


capplets/common/gtkrc-utils.c
```diff
@@ -60,16 +60,20 @@ gchar* gtkrc_find_named(const gchar* name)
...
-		{
-			g_free (path);
-			path = NULL;
-		}
-	}
+		const gchar * const * dirs = g_get_system_data_dirs();
+
+		if (dirs != NULL)
+			for (; !path && *dirs != NULL; ++dirs)
+                        {
```
jtojnar: Mixed spaces and tabs. Looks good otherwise.

romildo: > Mixed spaces and tabs. Looks good otherwise.

Fixed. Prefer tabs in this file, preserving the style already used.

PR #532, opened 2020-01-30
appearance: remove I/O error on window close

rbuj:
~/.xsession-errors 
```
I/O error : No such file or directory
```

### DIFF
--- a/capplets/appearance/mate-wp-xml.c
+++ b/capplets/appearance/mate-wp-xml.c
@@ -23,6 +23,7 @@
 #include <gio/gio.h>
 #include <string.h>
 #include <libxml/parser.h>
+#include <errno.h>
 
 static gboolean mate_wp_xml_get_bool(const xmlNode* parent, const char* prop_name)
 {
@@ -470,13 +471,10 @@ void mate_wp_xml_save_list(AppearanceData* data)
 	xmlNode* wallpaper;
 	//xmlNode* item;
 	GSList* list = NULL;
-	char* wpfile;
 
 	g_hash_table_foreach(data->wp_hash, (GHFunc) mate_wp_list_flatten, &list);
 	g_hash_table_destroy(data->wp_hash);
 	list = g_slist_reverse(list);
-
-		wpfile = g_build_filename(g_get_user_config_dir(), "mate", "backgrounds.xml", NULL);
 
 	xmlKeepBlanksDefault(0);
 
@@ -529,12 +527,24 @@ void mate_wp_xml_save_list(AppearanceData* data)
 		mate_wp_item_free(wpitem);
 	}
 
-	/* Guardamos el archivo, solo si hay nodos en <wallpapers> */
+	/* save the xml document, only if there are nodes in <wallpapers> */
 	if (xmlChildElementCount(root) > 0)
 	{
-		xmlSaveFormatFile(wpfile, wplist, 1);
+		g_autofree gchar *wpdir = NULL;
+		g_autofree gchar *wpfile = NULL;
+
+		wpdir = g_build_filename (g_get_user_config_dir(), "mate", NULL);
+		if (g_mkdir_with_parents (wpdir, 0700) == -1)
+		{
+			int errsv = errno;
+			g_warning ("failed, g_mkdir_with_parents(%s) failed: %s", wpdir, g_strerror (errsv));
+		}
+		else
+		{
+			wpfile = g_build_filename(wpdir, "backgrounds.xml", NULL);
+			xmlSaveFormatFile(wpfile, wplist, 1);
+		}
 	}
 
 	xmlFreeDoc(wplist);
-	g_free(wpfile);
 }


raveit65:
@rbuj 
With which fedora version you see the warning when closing mate-appearance-properties in ~/.xsession-errors ?

rbuj:
I saw it on Fedora rawhide, `~/.config/mate` folder doesn't exist.
```
$ find $HOME -name backgrounds.xml
```

rbuj:
Ubuntu 20.04 is also affected by this issue.
```shell
$ apt-cache show mate-control-center | grep Version
Version: 1.22.2-0ubuntu3
```

raveit65:
I can reproduce the warning with f30 if ~/.config/mate folder doesn't exist.

capplets/appearance/mate-wp-xml.c
```diff
@@ -529,12 +526,17 @@ void mate_wp_xml_save_list(AppearanceData* data)
 		mate_wp_item_free(wpitem);
 	}
 
-	/* Guardamos el archivo, solo si hay nodos en <wallpapers> */
+	/* save the xml document, only if there are nodes in <wallpapers> */
 	if (xmlChildElementCount(root) > 0)
 	{
+		char *wpdir = g_build_path (G_DIR_SEPARATOR_S, g_get_user_config_dir(), "mate", NULL);
```
b4n: I think you should use `g_build_filename()`, as it's the same but slightly simpler -- and you use it below.
```suggestion
		char *wpdir = g_build_filename (g_get_user_config_dir(), "mate", NULL);
```

capplets/appearance/mate-wp-xml.c
```diff
@@ -529,12 +526,17 @@ void mate_wp_xml_save_list(AppearanceData* data)
...
+		g_mkdir_with_parents (wpdir, 0700);
+		g_free (wpdir);
+
+		char *wpfile = g_build_filename(g_get_user_config_dir(), "mate", "backgrounds.xml", NULL);
```
b4n: I think you should reuse `wpdir` here and just add the filename to it instead of duplicating the directory part:
```suggestion
		char *wpfile = g_build_filename (wpdir, "backgrounds.xml", NULL);
```

capplets/appearance/mate-wp-xml.c
```diff
@@ -529,12 +526,17 @@ void mate_wp_xml_save_list(AppearanceData* data)
 		mate_wp_item_free(wpitem);
 	}
 
-	/* Guardamos el archivo, solo si hay nodos en <wallpapers> */
+	/* save the xml document, only if there are nodes in <wallpapers> */
 	if (xmlChildElementCount(root) > 0)
 	{
+		char *wpdir = g_build_path (G_DIR_SEPARATOR_S, g_get_user_config_dir(), "mate", NULL);
+		g_mkdir_with_parents (wpdir, 0700);
```
b4n: You should check the return value to know if the directory exists after the call, otherwise it might fail and still try and save to a non-existing location. [`g_mkdir_with_parents()` returns 0 if the directory exists after the call](https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-mkdir-with-parents).

capplets/appearance/mate-wp-xml.c
rbuj: Ok, I will change it ASAP.

REF:
https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-build-path

capplets/appearance/mate-wp-xml.c
rbuj: This could lead to conversion problems on some systems.

https://developer.gnome.org/glib/stable/glib-Character-Set-Conversion.html

capplets/appearance/mate-wp-xml.c
rbuj: If `mkdir -p $HOME/.config/mate` then I got `I/O error : No such file or directory`.

capplets/appearance/mate-wp-xml.c
rbuj: Ok, wip... it looks like g_get_user_config_dir()

b4n: > This could lead to conversion problems on some systems.

It should not, [`g_build_filename()`](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-build-filename) expect filename encoding arguments, so it should be alright.

raveit65:
> Looks good, but I couldn't actually test the new version yet.

Does a approval without testing make sense?
LOL

b4n:
I think it does given that I reviewed the code and my remarks have been addressed, but I can make it a "comment only" if you prefer.  I also mean to test this version, but I won't be able to before tomorrow at the soonest, probably the day after, and I don't want to block this because of my schedule.

PR #539, opened 2020-02-02
mate-theme-info: fix memory leak

rbuj:


sc0w:
I think the name of variable which needs to be freed need to mention into commit message (sound_theme_name)

rbuj:
@sc0w I detected the memory leaks using`-fsanitize=address` provided by libasan.

```shell
$ sudo dnf install libasan libasan-static libubsan libubsan-static liblsan liblsan-static
$ CFLAGS="-g -O0 -fsanitize=address,leak,undefined -fno-omit-frame-pointer -fno-common -Wformat -Werror=format-security -Werror=array-bounds" LDFLAGS="-fsanitize=address,undefined,leak -fno-omit-frame-pointer -fno-common" ./autogen.sh --prefix=/usr && make V=1 && sudo make install
```

sc0w:
@rbuj 

> I detected the memory leaks using-fsanitize=address provided by libasan.

ok, I think is good to add this check to travis

PR #568, opened 2020-05-18
[WIP] To fix : More than one of the same preferences window can be opened

### LINKED ISSUE #474: More than one of the same preferences window can be opened
#### Expected behaviour

You should only be able to open one instance of any particular preferences window. When you click to open a preferences window which is already open it should bring that window into focus, not open another of the same preferences.

#### Actual behaviour

When you open a preferences window that is already open, it opens a second window of the same preferences.

Open a preferences window twice, such as window properties or screensaver preferences.

#### MATE general version

1.20.1

#### Linux Distribution

Mint

tamplan:
Hi,

I want to fix #474 because i'm completly agree with the author.

I'm not familiar with edits of 'Makefile.am' and 'configure.ac' files.

I have changed the 'capplets/Makefile.am' file only for test purpose.

Based from the code on this page : https://developer.gnome.org/libwnck/stable/getting-started.html

The idea is to search a window with the same title of the capplet launched and close it if the window already exists.

Tested with success just with 'mate-about-me' to know if is the right way.

Can someone give me feedback please?

Thanks by advance.
Regards.

### DIFF
--- a/capplets/Makefile.am
+++ b/capplets/Makefile.am
@@ -1,15 +1,5 @@
 SUBDIRS = \
 	common \
-	accessibility \
-	appearance \
-	default-applications \
-	display \
-	keyboard \
-	keybindings \
-	mouse \
-	network \
-	time-admin \
-	windows \
 	about-me
 
 DIST_SUBDIRS = \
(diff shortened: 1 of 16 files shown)


tamplan:
I need to change the idea to transform all capplets to GtkApplication.

This PR must be deleted...

Issue #567, opened 2020-05-17
What is the /org/mate/desktop/background/draw-background key?


tamplan:
Hi,
I don't know where to post this question.

Can someone explain me why the /org/mate/desktop/background/draw-background key exist? And what is he function?

Is there a case where the background is not drawn?

Thanks by advance
Regards

raveit65:
With this key m-s-d draw the desktop when caja doesn't draw icons + background.

tamplan:
So, the background (the image) is always drawn:
- by Caja when the key is FALSE
- by m-s-d when the key is TRUE

In any case, the desktop page from m-c-c appearance capplet is always showed?

Is it right?

Thanks by advance.
Regards

raveit65:
```
[rave@mother ~]$ gsettings get org.mate.background draw-background
true
[rave@mother ~]$ gsettings get org.mate.background show-desktop-icons
true
[rave@mother ~]$
```
If both keys are `false` than you have a black desktop.
 `org.mate.background show-desktop-icons` is always prioritized.
When `org.mate.background show-desktop-icons` is  `true` it doesn't matter if `org.mate.background draw-background is set to true or false.

Honestly, i never tested if m-c-c shows the desktop page in this case :)
Try it out.

tamplan:
Where the 2 keys you mention are FALSE, the m-c-c shows the desktop page...

Let's go for a future PR...

Thanks you very much!

raveit65:
Please close topic!

PR #569, opened 2020-05-19
Apparence: draw-background key bug

tamplan:
When 'org.mate.background draw-background' is false, mate-appearance-properties should hide the background page.

Done with this PR.

When mate-appearance-properties is launched with argument '-p background', the app exit with no message.

Hop this helps.

Regards

### DIFF
--- a/capplets/Makefile.am
+++ b/capplets/Makefile.am
@@ -1,15 +1,5 @@
 SUBDIRS = \
 	common \
-	accessibility \
-	appearance \
-	default-applications \
-	display \
-	keyboard \
-	keybindings \
-	mouse \
-	network \
-	time-admin \
-	windows \
 	about-me
 
 DIST_SUBDIRS = \
--- a/capplets/about-me/Makefile.am
+++ b/capplets/about-me/Makefile.am
@@ -26,7 +26,8 @@ mate-about-me-resources.h mate-about-me-resources.c: org.mate.mcc.am.gresource.x
 
 bin_PROGRAMS = mate-about-me
 
-mate_about_me_LDADD = $(MATECC_CAPPLETS_LIBS)
+mate_about_me_LDADD = $(MATECC_CAPPLETS_LIBS)		\
+	$(LIBWNCK_LIBS)
 if HAVE_ACCOUNTSSERVICE
 mate_about_me_LDADD += $(ACCOUNTSSERVICE_LIBS)
 endif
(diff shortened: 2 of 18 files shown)


rbuj:
I could not reproduce your problem.

Captura de Pantalla 2020-05-19 a les 13 52 55

tamplan:
Have you test with the draw-background key to false? Without this PR.

tamplan:
Need to commit to remove g_print test message... :)

rbuj:
> Have you test with the draw-background key to false and without this PR?

This key determines whether msd can draw the background.
```
$ dconf dump /org/mate/desktop/background/
[/]
color-shading-type='solid'
draw-background=false
picture-filename=''
picture-options='wallpaper'
primary-color='rgb(39,88,126)'
secondary-color='rgb(98,127,90)'
show-desktop-icons=true
```
To prevent user to change the background you should lock background keys via dconf.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/desktop_migration_and_administration_guide/custom-default-values-system-settings

tamplan:
> To prevent user to change the background you should lock background keys via dconf.
I don't want to lock background keys.

Try with:
- set 'org.mate.background draw-background' to FALSE
- reboot
- you have got a blank screen
- launch mate-appearance-properties
- you have a page with you can change properties for a background the system don't show!

I think this is a bug, right?

This PR hide the background page if 'org.mate.background draw-background' is FALSE.

rbuj:
> > To prevent user to change the background you should lock background keys via dconf.
> > I don't want to lock background keys.
> 
> Try with:
> 
> * set 'org.mate.background draw-background' to FALSE
> * reboot
> * you have got a blank screen

No, the background was drawn as usual.

bg

tamplan:
Sorry, i forgot to give you more informations if i read this https://github.com/mate-desktop/mate-control-center/issues/567, it will better.

Please retry again with 'org.mate.background show-desktop-icons' set to FALSE too.

There is another bug but this PR don't matter of it.

'org.mate.background show-desktop-icons' to FALSE + 'org.mate.background draw-background' to FALSE give blank background on next reboot.

raveit65:
> This PR hide the background page if 'org.mate.background draw-background' is FALSE.

Why, this is a problem?
I don`t think that there is one user in the world who had both keys disable and use a black desktop :)
org.mate.background draw-background
org.mate.background show-desktop-icons

tamplan:
> > This PR hide the background page if 'org.mate.background draw-background' is FALSE.
> 
> Why, this is a problem?
> I don`t think that there is one user in the world who had both keys disable and use a black desktop :)
> org.mate.background draw-background
> org.mate.background show-desktop-icons

I think no too.
But, in my opinion, when a thing is disabled i don't see a reason to purpose to change properties about it and so, when org.mate.background draw-background is FALSE, there is no reason to purpose to user to change properties about a background not shown.

tamplan:
It seems i can purpose more than one PR at the same time even if there are on 2 different branches.
Travis can understand the previous PR i made is not related with this one...

PR #574, opened 2020-06-15
mate-window-properties.c: Update the button layout key for gtk-decorated windows

mtwebster:
mate-window-properties.c: Update the layout key for gtk client-decorate windows as well as marco's when the button layout is changed.

This is already implemented in mate-settings-daemon, and c-s-d windows
have become common enough that we should be updating these along with
wm-decorated windows.

### DIFF
--- a/capplets/windows/mate-window-properties.c
+++ b/capplets/windows/mate-window-properties.c
@@ -54,6 +54,9 @@
 #define MARCO_CENTER_NEW_WINDOWS_KEY "center-new-windows"
 #define MARCO_BUTTON_LAYOUT_KEY "button-layout"
 
+#define INTERFACE_SCHEMA "org.mate.interface"
+#define GTK_BUTTON_LAYOUT_KEY "gtk-decoration-layout"
+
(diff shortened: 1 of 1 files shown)


clefebvre:
I'd actually suggest adding another combo for this, specifically for CSD windows.

The reason for this is because the menu button is sometimes necessary in CSD apps, but redundant in normal titlebars (it shows nothing else than what the titlebar context menu already shows).

you might want :minimize,maximize,close in metacity, but menu:minimize,maximize,close in CSD.

clefebvre:
By the way, `:minimize,maximize,close` should be an option.

InfoLibre:
In Linux Mint MATE 20, buttons in Disks stay at left even when should be at right :
https://github.com/linuxmint/linuxmint/issues/280

raveit65:
@clefebvre @mtwebster 
Is this an active request or can we close it?

mbkma:
I think it can be closed, I implemented this code as part of #642

PR #588, opened 2020-09-01
keyboard-properties-a11y-notifications.ui: Fix a11y warnings

cwendling:
Fix a11y warnings by making the widget hierarchy more semantic, as well as making it saner (no more space-filled labels to create margins).  This rewrites a part of the dialog, but the only visual change is a slight difference in title margins due to using a stock `GtkFrame`.

mcc-ui1
Before

mcc-ui2
After

### DIFF
--- a/capplets/keyboard/mate-keyboard-properties-a11y-notifications.ui
+++ b/capplets/keyboard/mate-keyboard-properties-a11y-notifications.ui
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.0 -->
+<!-- Generated with glade 3.22.2 -->
 <interface>
   <requires lib="gtk+" version="3.14"/>
   <!-- interface-license-type gplv2 -->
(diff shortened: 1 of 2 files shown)


raveit65:
Is it OK to remove all the GDK_Event_MASK properties ?
Eg.
```
<property name="events">GDK_POINTER_MOTION_MASK | GDK_POINTER_MOTION_HINT_MASK | GDK_BUTTON_PRESS_MASK | GDK_BUTTON_RELEASE_MASK</property>
```

cwendling:
@raveit65 FAIACT, yes: those should be for the various widgets to catch motion and button press/release; but there's nothing special the code does, so the defaults for each widget should be fine.  Also, I can't see any behavioral change, and all expected interaction work (mouse clicks, keyboard shortcuts).

raveit65:
> FAIACT

Typo?
No idea what does this mean

cwendling:
Oops, yes, I meant AFAICT (which is short for *as far as I can tell*, just in case), sorry.

raveit65 (approved):
LGTM,
UI window comes up fine

PR #591, opened 2020-10-16
Appearance applet: reduce and simplify the code

rbuj:
It also fixes the stock deprecations in the UI file.

### DIFF
--- a/capplets/appearance/appearance.h
+++ b/capplets/appearance/appearance.h
@@ -137,3 +137,4 @@ typedef struct {
 } AppearanceData;
 
 #define appearance_capplet_get_widget(x, y) (GtkWidget*) gtk_builder_get_object(x->ui, y)
+#define GET_WINDOW(x) (GTK_WINDOW (gtk_builder_get_object (data->ui, (x))))
--- a/capplets/appearance/appearance-desktop.c
+++ b/capplets/appearance/appearance-desktop.c
@@ -613,7 +613,7 @@ wp_create_filechooser (AppearanceData *data)
 
   data->wp_filesel = GTK_FILE_CHOOSER (
                      gtk_file_chooser_dialog_new (_("Add Wallpaper"),
-                     GTK_WINDOW (appearance_capplet_get_widget (data, "appearance_window")),
+                     GET_WINDOW ("appearance_window"),
                      GTK_FILE_CHOOSER_ACTION_OPEN,
                      "gtk-cancel",
                      GTK_RESPONSE_CANCEL,
--- a/capplets/appearance/appearance-desktop.c
+++ b/capplets/appearance/appearance-desktop.c
@@ -1328,8 +1328,6 @@ desktop_init (AppearanceData *data,
                     (GCallback) wp_scolor_changed, data);
 
   add_button = appearance_capplet_get_widget (data, "wp_add_button");
-  gtk_button_set_image (GTK_BUTTON (add_button),
-                        gtk_image_new_from_icon_name ("list-add", GTK_ICON_SIZE_BUTTON));
 
   g_signal_connect (add_button, "clicked",
                     (GCallback) wp_file_open_dialog, data);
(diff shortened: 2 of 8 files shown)


raveit65 (commented):
LGTM,
builds and runs fine.

raveit65 (approved):

selectiveduplicate (approved):
Looks good.

capplets/appearance/appearance.h
```diff
@@ -137,3 +137,4 @@ typedef struct {
 } AppearanceData;
 
 #define appearance_capplet_get_widget(x, y) (GtkWidget*) gtk_builder_get_object(x->ui, y)
+#define GET_WINDOW(x) (GTK_WINDOW (gtk_builder_get_object (data->ui, (x))))
```
cwendling, line 140: I don't like macros that use local variables implicitly, unless they are highly specific and defined right next to the code using them.  Here, I'd really prefer seeing `data` be a parameter -- which basically makes it `GTK_WINDOW (appearance_capplet_get_widget (x, y))`.

The biggest reason why I think it's better is that implicit use of local variables make it a lot harder to understand when reading, and a lot easier to debug would anything go wrong (e.g. a `undefined variable 'data'` would be cryptic for the caller, so would `'data' does not have a member named 'ui'` or such).  It also makes it slightly harder to use, as if you happen to already have a variable `data` you need to refactor your code to be able to use that macro.

rbuj: I don't think it's necessary to add a new variable in order to predict a future refactoring, such as builder on the sample below:

```
GtkBuilder *builder = data->ui;
...
... GET_WINDOW (label) ...
...
#define GET_WINDOW(x) (GTK_WINDOW (gtk_builder_get_object (builder, (x))))
```

cwendling: I don't think we need a new magic variable, I think we need an additional parameter to the `GET_WINDOW()` macro:
```c
#define GET_WINDOW(appearance_data, y) (GTK_WINDOW (appearance_capplet_get_widget (appearance_data, y)))
```
and call it as `GET_WINDOW (data, "some member")`.

rbuj: You do not need to add a new parameter, if it always takes the same value.

cwendling: My point is that it adds magic to that macro, and requirements on the environment in which it is called.  Both of these make it harder to use, and prone to errors or headaches in the future.  And I don't think saving typing 6 characters (`data, `) is worth this kind of trouble.
In other words, I don't think having one parameter less is not worth requiring the caller to have a variable with a specific name and semantic defined to be able to use this macro.

However, feel free to ignore my remarks, but basically my experience is "magic" in macros leads sooner or later to problems, unless it is *very* clear and well understood by all parties (callers and callee).

rbuj: In the case you describe you would no longer need to use the GET_WINDOW macro and you would use the value replaced by the processor directly.

cwendling: Well, using the macro would save the caller from some typing, which is what I though your goal was.  Instead of `(GTK_WINDOW (appearance_capplet_get_widget (data, "something")))`, the caller would type `GET_WINDOW (data, "something")`.

lukefromdc (approved):
We have a LOT of macros removed here, but this did finish building and at least appear to have the same build warning. Runtime behavior seems to be the same as master, so at least in Debian and at least with gcc9 this works.

Two issues are the same with this and with master: one is the torrent of 
`../gdk/x11/gdkwindow-x11.c:5633 drawable is not a native X11 window`
warnings when running mate-appearance-properties from a terminal window,
and the other is inability to delete the newly made "test" theme saved in testingh the ability to save a theme. Neither issue has anything to do with this PR, but you might see them when testing it so I mention them here.

lukefromdc:
Is this ready to go? we have three approving reviews but some unresolved conversations

raveit65:
@cwendling 
Is your concern a "NO GO" or can this be merged?

cwendling:
No, I won't block this.  My concerns haven't been addressed, but you can see them as diverging opinions rather than an issue proper.

PR #592, opened 2020-10-17
appearance: Fix preview in Add Wallpaper dialog

rbuj:
Screenshot at 2020-10-17 13-13-31


lukefromdc (approved):
Builds and runs fine. This makes a preview of the image appear in the filechooser dialog when adding another image to the cached list of wallpaper images. In current master you get a question mark (at least w my icon theme) in the center of the right hand margin of the filechooser. With this, a preview of the selected image is shown. 

Looks like the previous code failed due to trying to change the value of a constant

raveit65 (approved):

raveit65 (approved):
LGTM,
works well.

PR #600, opened 2020-11-24
mouse: size group

rbuj:


raveit65 (approved):
LGTM,
builds and runs fine.
Dialog looks like before.

PR #608, opened 2020-11-27
Revisit windows applet, add allow-top-tiling option

mbkma:
The design is more similar to the original now.
Adds allow-top-tiling option, which is insensitive if use-reduced-resources is true or if allow-tiling (Enable window tiling) is false.

Screenshot at 2020-11-27 12-22-00
Screenshot at 2020-11-27 12-22-20
Screenshot at 2020-11-27 12-22-29


rbuj (approved):
LGTM

raveit65 (approved):
LGTM,
thanks for adding the setting `allow-top-tiling`.

PR #614, opened 2021-01-23
Remove mate-screensaver support

rbuj:
Related mate-desktop/mate-screensaver#243


raveit65:
We can reopen it if we really want replace mate-screensaver.
Personal i prefer to get proper licensing for ms.

PR #619, opened 2021-02-13
i18n: fix an error that prevents showing the wallpaper name

rbuj:
Screenshot at 2021-02-13 10-57-29


raveit65:
What is the error?
m-a-p

rbuj:
Yes, I remember there was an error in the way the xml file was loaded, maybe this error no longer exists now.

If the xml data aren't aligned with the tree traversal algorithm, the parsing of the xml file fails as it also parses the element's attributes on each iteration. This commit only parses the xml elements, which are sibling xml nodes.

raveit65 (approved):
LGTM,
works like before.

PR #624, opened 2021-02-22
windows: missing include mate-metacity-support.h

rbuj:
Fix the warning below:
```
mate-metacity-support.c:30:1: warning: no previous prototype for ‘mate_metacity_config_tool’ [-Wmissing-prototypes]
```

### DIFF
--- a/capplets/windows/mate-metacity-support.c
+++ b/capplets/windows/mate-metacity-support.c
@@ -23,6 +23,8 @@
 #include <gio/gio.h>
 #include <gtk/gtk.h>
 
+#include "mate-metacity-support.h"
+
 #define METACITY_SCHEMA "org.gnome.metacity"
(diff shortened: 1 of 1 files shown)


zhuyaliang (approved):

zhuyaliang:
Compilation warning removed successfully

lukefromdc (approved):
File builds clean without warnings. Tested this with metacity, note that with the version of metacity I have installed the only option is to toggle compositing on or off. We have some opti

That always reverts to compositing ON, my guess is metacity has removed support for disabling composting. Same behavior in dconf-editor proves this is metacity and not mate-control-center's doing. Dconf-editor brings up a few more metacity options to turn alt-tab thumbnails on or off, placement mode, and turning edge tiling on or off. Supporting these would be new commits, if supported in the past these are different issues than what this fixes.

In other words, this change itself looks good to go.

PR #629, opened 2021-02-24
marco-window-manager: get XDG_DATA_DIRS (code portability)

rbuj:
based on marco code https://github.com/mate-desktop/marco/blob/7ab70915e515337e9900809ddaf404f8d57622ec/src/ui/theme-parser.c#L4395-L4418


lukefromdc (approved):
Builds and runs without issue on Debian Unstable. Anything relating to portability would ideally be tested somewhere that matters (BSD?) but I doubt that will happen before the 1.26 release due to team size

zhuyaliang (approved):

PR #630, opened 2021-02-24
Line comments // are not allowed (code style)

rbuj:
Reported by eclipse IDE


lukefromdc (approved):
Builds fine, no typos in here to break the build as confirmed by a quick test build.
Is this something we normally enforce in our coding style? I can't speak to that and don't use Eclipse so never seen that warning

zhuyaliang (approved):
I haven't used the eclipse IDE, but it's a good thing to use a unified line comment format

PR #631, opened 2021-02-24
bookmark-agent: Function 'bookmark_agent_new' used prior to declaration

rbuj:
Reported by NetBeans IDE


lukefromdc (approved):
This builds and runs fine, we just move a block of code down a bit.

zhuyaliang (approved):

PR #638, opened 2021-03-25
appearance: Type column in GtkFileChooser requires gtk 3.24.12

### LINKED ISSUE #637: Commit 844db92 (appearance: Add file chooser settings - Nov 29, 2020) breaks Mate Control Center if GTK3 < 3.24.12
#### Expected behaviour
Mate Control Center launch, and is usable as expected

#### Actual behaviour
Mate Control Center (as well as Appearance Properties binary) does not launch on CentOS 8.3 at all, the following is thrown out if launched from terminal:
(mate-appearance-properties): GLib-GIO-ERROR **: Settings schema 'org.gtk.Settings.FileChooser' does not contain a key named 'show-type-column'

launch 'mate-control-center' binary or 'mate-appearance-properties' binary from a terminal on any system with gtk3 version less than 3.24.12

#### MATE general version
1.25.1 (git commits after 844db92, by today)

#### Package version
Custom build on CentOS 8.3 / SPEC file based on Fedora 1.24 SRPM

#### Linux Distribution
CentOS 8

#### Link to bugreport of your Distribution (requirement)
-

As much as I've been able to find, [commit 844db921aeed3e987f3be0f4e7e0b3a9703612ac](https://github.com/mate-desktop/mate-control-center/commit/844db921aeed3e987f3be0f4e7e0b3a9703612ac) introduced "File Chooser" settings (in Appearance Preferences / Interface tab), among which "Show type column" checkbox / setting is not available in GTK3 before 3.24.12.

I've created a patch which fixes the crash:
[patch-1.diff.txt](https://github.com/mate-desktop/mate-control-center/files/6198277/patch-1.diff.txt)

With this patch, I was able to launch both mate-control-center and mate-appearance-properties, but the problematic setting was still available (even if not useful), what could be misleading.

I've therfore added a patch for glade appearance.ui file, which shall be applied conditionally (if gtk3 < 3.24.12 only - to inactivate / grey out this checkbox / setting):
[patch-2.diff.txt](https://github.com/mate-desktop/mate-control-center/files/6198278/patch-2.diff.txt)

It would maybe be better to inactivate / grey out this checkbox / setting at runtime (if gtk3 < 3.24.12 / setting not available), or hide it, but I don't want to mess with the code more than really necessary (and, I feel to be safe - I'm not a coding expert at all...).

rbuj:
closes #637

### DIFF
--- a/capplets/appearance/appearance-main.c
+++ b/capplets/appearance/appearance-main.c
@@ -190,6 +190,10 @@ main (int argc, char **argv)
   capplet_set_icon (w, "preferences-desktop-theme");
   gtk_widget_show_all (w);
 
+#if !GTK_CHECK_VERSION(3, 24, 12)
+  gtk_widget_hide (GTK_WIDGET (gtk_builder_get_object (data->ui, "filechooser_type_column_toggle")));
+#endif /* !GTK_CHECK_VERSION(3, 24, 12) */
+
   g_signal_connect_after (w, "response",
                           (GCallback) main_window_response, data);
 
--- a/capplets/appearance/appearance-ui.c
+++ b/capplets/appearance/appearance-ui.c
@@ -114,11 +114,13 @@ ui_init (AppearanceData *data)
                      "active",
                      G_SETTINGS_BIND_DEFAULT);
 
+#if GTK_CHECK_VERSION(3, 24, 12)
     g_settings_bind (data->filechooser_settings,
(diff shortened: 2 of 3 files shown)


zhuyaliang (approved):

raveit65:
@L-U-T-i
Does this works for you on CentOS-8.3?

raveit65 (approved):
Works fine with gtk3-3.24.28-1.fc34.x86_64

PR #646, opened 2021-04-26
undo windows-applet: ... remove reduced-resources option

rbuj:
reduced-resources is an active preference in marco


capplets/windows/mate-window-properties.c
```diff
@@ -429,13 +435,19 @@ main (int argc, char **argv)
                      "active",
                      G_SETTINGS_BIND_DEFAULT);
 
-    /* Compositing Manager */
+    /* Composition Manager */
```
rbuj, line 438: Composition has another meaning that is unrelated to this domain.
https://en.wikipedia.org/wiki/Compositing_window_manager

PR #650, opened 2021-07-18
Fix compilation warning

zhuyaliang:
Fixed some compilation warnings that there were ``` no previous prototypes [- wmissing prototypes]```
We still have a lot to do. There are too many outdated APIs that we need to replace


vkareh (approved):
LGTM

PR #651, opened 2021-07-24
Replace the deprecated GtkTable with GtkGrid

zhuyaliang:
Reduce compilation warnings 
Replace the deprecated GtkTable with GtkGrid
The display effect is the same as before, and there is no warning at runtime

### DIFF
--- a/libslab/app-resizer.c
+++ b/libslab/app-resizer.c
@@ -62,33 +62,22 @@ remove_container_entries (GtkContainer * widget)
 }
 
 static void
-resize_table (GtkTable * table, gint columns, GList * launcher_list)
+resize_table (AppResizer *widget, GtkGrid * table, gint columns)
 {
-	float rows, remainder;
-
 	remove_container_entries (GTK_CONTAINER (table));
-
-	rows = ((float) g_list_length (launcher_list)) / (float) columns;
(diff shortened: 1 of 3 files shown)


libslab/app-resizer.c
```diff
@@ -223,11 +211,8 @@ app_resizer_size_allocate (GtkWidget * widget, GtkAllocation * allocation)
 	GtkRequisition other_requisiton;
 	gtk_widget_get_preferred_size (GTK_WIDGET (resizer->cached_tables_list->data), &other_requisiton, NULL);
 
-	useable_area =
-		allocation->width - (child_requisition.width -
-		other_requisiton.width);
 	new_num_cols =
-		relayout_tables_if_needed (APP_RESIZER (resizer), useable_area,
+		relayout_tables_if_needed (APP_RESIZER (resizer), allocation->width,
```
zhuyaliang, line 215: Because we removed ```gtk_table_resize (table, (int) rows, columns);```,  ```child_requisition.width``` should not be used,```child_requisition.width``` will always change with the increase of components.

raveit65 (approved):
LGTM,
builds without new warnings and m-c-c UI runs fine.

vkareh (approved):
works for me, grid layout reflows nicely.

PR #654, opened 2021-07-24
Remove useless and discarded gtk_button_released

zhuyaliang:
```gtk_button_released``` has been deprecated, and we don't need to send signals manually


zhuyaliang:
I haven't found a specific use for this signal, but in order to be consistent with the past, I will use ```g_signal_emit_by_name ```replace ```button-release-event```

vkareh (approved):
Works well for me

raveit65 (approved):
LGTM

Issue #655, opened 2021-08-09
mate-control-center 1.26.0 build fails if glib is < 2.66


darkshram:
Because commit 71c7563e708128f80d6d60447cdb0430c5414626, mate-control-center 1.26.0 only builds with glib >= 2.66. According to GTK documentation, [g_bookmark_file_set_modified_date_time](https://docs.gtk.org/glib/method.BookmarkFile.get_modified_date_time.html) is available only in glib >= 2.66.

raveit65:
Which distro is using glib <=2.66 ?

darkshram:
> Which distro is using glib <=2.66 ?

[https://pkgs.org/search/?q=glib](https://pkgs.org/search/?q=glib)

- openSUSE Leap 15.3
- Slackware 14.2
- Ubuntu 20.04 LTS (Focal Fossa)
- RHEL 8
- CentOS 8
- AlmaLinux 8 
- RockyLinux 8
- ALDOS 1.4
- And any Linux distribution with LTS releases. Most of these use glib 2.58 or 2.62.

I had to build my packages reverting 71c7563e708128f80d6d60447cdb0430c5414626.

raveit65:
I am in doubt that all those distros will get an update by distrospackage maintainers.
Eg. Rhel8 / Centos8 doesn't ship  MATE by official builds or by fedora repos (me).
And I don't think that @flexiondotorg  will ever update Ubuntu 20.04 LTS (Focal Fossa) with mate-1.26.
In result i prefer to pump glib requires of m-c-c.

raveit65:
Btw. why not open a PR which use old an new code with a glib check?

thesquash:
> I don't think that @flexiondotorg will ever update Ubuntu 20.04 LTS (Focal Fossa) with mate-1.26.

You're right, @raveit65.  Let's face it:  Ubuntu LTS releases have *never* (that I know of) upgraded desktop environment packages much beyond the initial release of the LTS.  It's simply the antithesis to the LTS release idea to include a newer major version of the MATE desktop in an already-produced LTS release.

The closest that we have thus far is [this unstable MATE PPA](https://launchpad.net/%7Eubuntu-mate-bot/+archive/ubuntu/daily-brew), which targets Ubuntu MATE 21.04 Hirsute Hippo -- which is not an LTS release.

raveit65:
Again, why not opening a PR with
```
#if GLIB_CHECK_VERSION(2,66,0)
<new code>
[...]

PR #658, opened 2021-08-24
Set required Glibc to v2.66

flexiondotorg:
g_bookmark_file_set_modified_date_time is used in libslab/bookmark-agent.c which sets the last time the bookmark for uri was last modified.

g_bookmark_file_set_modified_date_time is only available in glibc 2.66 and newer.

https://docs.gtk.org/glib/struct.BookmarkFile.html


flexiondotorg:
Closing this, as discussed here:

  * https://github.com/mate-desktop/mate-control-center/issues/655

PR #663, opened 2021-09-21
Remove redundant i18n strings

yarons:
These specific functions can only cause mess when translated, it's better to keep the original terms.


raveit65 (approved):
LGTM,
thank you.

PR #665, opened 2021-10-12
Include audio mixers in hardware category

rbuj:
Rule: AudioVideo AND Mixer AND (NOT System)
- [AudioVideo](https://specifications.freedesktop.org/menu-spec/latest/apa.html): Main Category
- [Mixer](https://specifications.freedesktop.org/menu-spec/latest/apas02.html): Additional Categories
- [System](https://specifications.freedesktop.org/menu-spec/latest/apa.html): Main Category

### DIFF
--- a/shell/matecc.menu
+++ b/shell/matecc.menu
@@ -76,6 +76,11 @@
         <Category>HardwareSettings</Category>
         <Not><Category>System</Category></Not>
       </And>
+      <And>
+        <Category>AudioVideo</Category>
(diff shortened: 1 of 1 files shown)


rbuj:
I have closed the PR as I believe there is a simpler approach. See https://github.com/mate-desktop/mate-media/pull/176

raveit65:
I would really prefer a solution without moving the application from audio-tools at main-panel-menu.

PR #668, opened 2021-10-22
Fix memory leaks II

rbuj:


mbkma (approved):
see https://developer-old.gnome.org/glib/stable/glib-Commandline-option-parser.html#g-option-context-new

mbkma:
Anyone likes to review this PR?

raveit65:
> Anyone likes to review this PR?

@rbuj 
Sorry, i can't review more PRs because credits by Travis CI was running out again.
You should ask for more credits for yourself.

lukefromdc (approved):
Both the keyboard and mouse capplets work fine with this. I am assuming they would both crash on startup if ` g_option_context_free() ` was being called prematurely.

Note that I had to build with --disable-nls due to a missing sv.po file no doubt inherited from master given that no changes are made to po files in this PR

PR #670, opened 2021-10-22
common: fix memory leak detected by scan-build

rbuj:


mbkma (approved):
LGTM

lukefromdc (approved):
mate-theme-info.c builds with no warnings at all with this and my default build options. Changing themes with this installed seems to work fine

PR #674, opened 2021-10-26
bookmark-agent: remove whitespaces before #if, #else, and #endif

rbuj:
Test:
```
grep -R -w "^[[:space:]]*#if\|^[[:space:]]*#else\|[[:space:]]*#endif" --include *.c
```


zhuyaliang (approved):
Build normal

lukefromdc (approved):
bookmark-agent.lo builds with no warnings with this. Not sure what the bookmark agent does, but didn't notice any behavioral changes. This is expected as removing these whitespaces should have no effect on the built binaries

PR #679, opened 2021-11-01
bookmark-agent: fix backward compatibility with glib < 2.66

rbuj:
See #659


lukefromdc (approved):
Builds and runs fine with locally compiled glib 2.69.0 but needs testing with the older versions as well

L-U-T-i:
Just built with glib 2.56.4 (Rocky Linux 8.5), no issues.

lukefromdc:
OK, this has passed tests with both newer and older glib so merging

PR #680, opened 2021-11-01
bookmark-agent: fix memory leak

rbuj:


lukefromdc (approved):
Builds clean, no warnings from this file at all. No problems running the built mate-control-center(edit: retested w full install) so this looks good. Test is with Debian unstable and locally compiled glib 2.69.0

cwendling (changes requested):
IIUC `item->mtime` is only assigned to the value returned from [`get_modified_date_time()`](https://docs.gtk.org/glib/method.BookmarkFile.get_modified_date_time.html) [line 329](https://github.com/mate-desktop/mate-control-center/pull/680/files#diff-290d9677a0b84fb67ecb89c98deb34a5cfa245cfc23c18b53710f5a25f9f401cL329), which as the documentation states, does not transfer ownership of the returned value.  So it must not be freed.

What is your evidence that there is a memory leak here?

---
~~OT: looks like this code is supposed to support GLib < 2.66 where [`get_modified_date_time()`](https://docs.gtk.org/glib/method.BookmarkFile.get_modified_date_time.html) is not present, but [looking at the header file](https://github.com/mate-desktop/mate-control-center/blob/365ce438e6fe334f43f12f16f53a134b8b6f97df/libslab/bookmark-agent.h#L44) suggests it only works with >= 2.66 as there is no equivalent checks for making `mtime` a `time_t` on earlier versions.  Technically it could work because a `GDateTime*` is probably large enough to hold a `time_t`, but meh.  This however is totally off-topic, an I just noticed because I wondered why the version check in this patch.~~ I missed #679 which addresses the issue.

PR #683, opened 2021-11-23
Remove build warning -Wbad-function-cast

rbuj:


zhuyaliang (approved):
Warning fixed

lukefromdc (approved):
Changing themes and displaying fonts within mate-appearance-properties->fonts and clicking on one of the font options work as expected with this

PR #692, opened 2022-02-08
mate-theme-info: code formatting

rbuj:


lukefromdc (approved):
Builds and runs fine, one typo and a broken build would be expected. This looks good and should build the same code unless I missed some intended functional change in all the modified text. Changing themes works as expected.

Note that the string of warnings:
`(mate-appearance-properties:108856): Gdk-WARNING **: 23:21:00.660: gdkwindow-x11.c:5653 drawable is not a native X11 window`
you get running mate-control-center ->change theme from terminal are the same in master so they have nothing to do with this PR

lukefromdc:
Thought this had another approval, hope we don't get a surprise forcing a revert

PR #698, opened 2022-08-31
mouse: feature pointer size control

monnerat:
This adds a mouse/touchpad pointer size control scale to both pages of the mouse properties dialog.


capplets/mouse/mate-mouse-properties.ui
```diff
@@ -610,6 +617,153 @@ Author: Wolfgang Ulbrich
...
+                                    <child>
+                                      <object class="GtkScale" id="pointer_size_scale">
+                                        <property name="visible">True</property>
+                                        <property name="can-focus">True</property>
+                                        <property name="adjustment">adjustment8</property>
+                                        <property name="digits">0</property>
+                                        <property name="draw-value">False</property>
+                                        <child internal-child="accessible">
+                                          <object class="AtkObject" id="pointer_size_scale-atkobject">
+                                            <property name="AtkObject::accessible-description" translatable="yes">Cursor blinks speed</property>
```
cwendling, line 710: that looks off

capplets/mouse/mate-mouse-properties.ui
```diff
@@ -1572,6 +1726,153 @@ Author: Wolfgang Ulbrich
...
+                                    <child>
+                                      <object class="GtkScale" id="touchpad_pointer_size_scale">
+                                        <property name="visible">True</property>
+                                        <property name="can-focus">True</property>
+                                        <property name="adjustment">adjustment8</property>
+                                        <property name="digits">0</property>
+                                        <property name="draw-value">False</property>
+                                        <child internal-child="accessible">
+                                          <object class="AtkObject" id="touchpad_pointer_size_scale-atkobject">
+                                            <property name="AtkObject::accessible-description" translatable="yes">Cursor blinks speed</property>
```
cwendling, line 1819: ditto

cwendling (dismissed):
Do we really want this?  If you can control the size of the pointer there, why not the aspect/theme?

IMO if anything there should be a button opening the pointer theme options which include both, or redesign things to have everything in one place (a third tab with the theme, if really we want it there?  but then there is the same problem in the theme dialog which would have to link here).

Also, I don't like the idea there are two sliders that control the same thing: it *looks* like you could control the size of the mouse and touchpad pointers separately with this, but it's not the case (and is not really a thing anyway).

Finally, this duplicates all the details of the slider (I didn't check if they were consistent), especially min/max values, which can be annoying to maintain and keep in sync everywhere.

monnerat:
> Do we really want this? If you can control the size of the pointer there, why not the aspect/theme?

From an UI designer point of view, you're right.
On a normal desktop user standpoint, this is natural to configure that as a mouse option.
I often had to help people that don't think to look in theme config for this option. This is not obvious for most of them: in their minds, the pointer is tightly bound to the mouse hardware.

> IMO if anything there should be a button opening the pointer theme options which include both, or redesign things to have everything in one place (a third tab with the theme, if really we want it there? but then there is the same problem in the theme dialog which would have to link here).

This could be a possibility but I don't want to make a revolution! Gnome people already did and that was the reason for the Mate fork!
I addition, both these suggestions are much heavier in terms of changes.

> Also, I don't like the idea there are two sliders that control the same thing: it looks like you could control the size of the mouse and touchpad pointers separately with this, but it's not the case (and is not really a thing anyway).

I just mimicked what is done for the "Pointer Speed" option, that already duplicates controls.

> Finally, this duplicates all the details of the slider (I didn't check if they were consistent), especially min/max values, which can be annoying to maintain and keep in sync everywhere.

Min/max/step values are taken from a GtkAdjustment that is shared between both scales.

capplets/mouse/mate-mouse-properties.ui
monnerat: Yes,you're right: this is a leftover after a cut+paste from another widget... that is wrong too ! This comes originally from the keyboard ui.
I'll remove these (from other widgets too).

capplets/mouse/mate-mouse-properties.ui
monnerat: Same

capplets/mouse/mate-mouse-properties.ui
monnerat: Fix pushed.

mbkma:
can't you already configure pointer size in AppearancePreferences->Theme->Customise->Pointer->Size ?

monnerat:
> can't you already configure pointer size in AppearancePreferences->Theme->Customise->Pointer->Size ?

Yes, you can.
But the path is so complicated and unnatural that normal users never find it.
From what I've noticed, they always expect it in the mouse config.
A redundant place for that doesn't hurt and avoids requests to IT people.

If you're seeking for complex ways to do it, you can also use dconf-editor :-)

cwendling:
what about a button/link "Go to cursor appearance preferences" that would pop up the appearance prefs dialog at the right tab?

monnerat:
> what about a button/link "Go to cursor appearance preferences" that would pop up the appearance prefs dialog at the right tab?

Could be an alternate approach if the proper page can be accessed.
But I don't think it's currently possible as you can only use `--show-page=[theme|background|fonts|interface]` and we have to target page `Pointer` from the subdialog `Customize...` that can only be reached via a button in the `Appearance` `Theme` page..

In addition, `mate-at-properties` links to mouse config (among others) but not to the appearance dialog.

cwendling:
That could probably be augmented to include the missing bits, it's not really a problem if it needs to pop another dialog on top of it, it's quite easy to do it.  But it has to be done indeed, it's not already there.

monnerat:
I don't think it would keep it simple, although logic from a UI designer standpoint:

- Additional piggy-back code in `mate-appearance-properties`
- More windows open for the user, even the `Theme` dialog that is not really needed.

lukefromdc:
What is the status of this?

luigifab:
The idea of the link is a good solution that open the `mate-appearance-properties` dialog at the right place.

It could be nice to add it also for `mate-window-properties`.
I suggest a new tab called Theme for each dialog.

vkareh:
The `mate-mouse-properties` capplet is arguably entirely about the device itself (*mouse*, *touchpad*, etc.) and its behavior. Each tab refers to a distinct hardware device.

Size is a theme/appearance concern of the *pointer* (independent of the device). The configuration affects the visual appearance of the pointer regardless of which device controls it.

There is some overlap ("mouse" controls the "pointer") but not enough (in my mind) to justify duplicating the pointer theme functionality in the mouse/touchpad configuration dialog.

The only thing that makes sense is having an "Appearance" button next to the Help button at the bottom of the window (outside the tab container). This button should open the Pointer customization window (this needs to be programmed, as it's not possible today). Even then this is awkward, since that window also has tabs for gtk, window decorations, and icons, unless it can be isolated somehow.

Tempted to close this and wait to see if someone wants to work on that suggestion.

*Edit:* Another idea would be to have the size slider outside the tab container, maybe in the space between the Help and Close buttons. Maybe this is awkward too, but it may be a valid solution. Also, not that we need to follow GNOME's lead, but they moved the cursor size as part of the accessibility controls.

PR #704, opened 2023-05-24
windows: Fix UI accessibility

### LINKED ISSUE #703: window-properties.ui labels need an accessible relation pointing to the group they label
Steps to reproduce:
1. Launch Orca and mate-window-properties
2. Tab among the widgets

Expected results: Orca would announce each new section of controls.

Actual results: Orca does not announce each new section of controls. Often the widget labels are sufficient to understand what the widget is for. But when you give focus to "Alt", all Orca says is "Alt". Orca doesn't say "Movement Key" or "To move a window, press-and-hold this key then grab the window".

The reason Orca doesn't announce the labels is that it has no idea what those labels are associated with. There are ATK relations which you can use to fix this.

Related aside: Orca does announce the "To move a window...." label text when the dialog first appears. That's because Orca is looking for long static text that might be helpful information that should be presented for newly-opened dialogs. Because this label doesn't have an accessible relation pointing to the radio  button group, and because it's sufficiently long, it passes Orca's heuristic. While I (Orca maintainer) can try to give Orca more smarts to eliminate that initial chattiness, it's not going to solve the problem that tabbing to the "Alt" radio button doesn't result in the user being told what choosing "Alt" will do. The accessible relation needs to be added. And once you add it, that initial chattiness will go away on its own.

cc @cwendling

cwendling:
Use real frames for a semantic UI so GTK and screen readers can set up and figure out accessibility relationships between the frame label and the content automatically.

Also properly add relationships between the movement keys descriptive label and the key radio items so a screen reader can pick it up.  As a bonus, it also prevents Orca from erroneously announcing the label when the window comes up because it thinks this long orphaned label is probably a general description for the window.

Fixes #703.

### DIFF
--- a/capplets/windows/mate-window-properties.c
+++ b/capplets/windows/mate-window-properties.c
@@ -113,6 +113,7 @@ static GtkWidget *focus_mode_mouse_checkbutton;
 static GtkWidget *autoraise_checkbutton;
 static GtkWidget *autoraise_delay_spinbutton;
 static GtkWidget *autoraise_delay_hbox;
+static GtkWidget *movement_description_label;
 static GtkWidget *alt_click_vbox;
 
 /* Placement */
@@ -402,6 +403,7 @@ main (int argc, char **argv)
     autoraise_delay_hbox = GET_WIDGET ("autoraise_delay_hbox");
     autoraise_checkbutton = GET_WIDGET ("autoraise_checkbutton");
     autoraise_delay_spinbutton = GET_WIDGET ("autoraise_delay_spinbutton");
+    movement_description_label = GET_WIDGET ("movement_description_label");
     alt_click_vbox = GET_WIDGET ("alt_click_vbox");
 
     /* Placement */
@@ -543,6 +545,7 @@ reload_mouse_modifiers (void)
     gboolean have_super;
     int min_keycode, max_keycode;
     int mod_meta, mod_super, mod_hyper;
+    AtkObject *label_accessible;
 
     XDisplayKeycodes (gdk_x11_display_get_xdisplay(gdk_display_get_default()),
                       &min_keycode,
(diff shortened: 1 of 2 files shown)


zhuyaliang (approved):
Compilation and runing well

raveit65 (approved):
LGTM,
not tested with orca but the behavior tab is marked as focused after application start with Blue-Menta theme and i could tab through the application.
I guess this is the goal of this PR.
Maybe someone with having orca running could test this PR.

cwendling:
> LGTM, not tested with orca but the behavior tab is marked as focused after application start with Blue-Menta theme and i could tab through the application. I guess this is the goal of this PR.

No, tabbing behavior was already OK, the goal of this PR is to make Orca able to understand what section labels (the bold text) relate to.  With this PR, Orca will use those labels to announce the settings groups, just like you visually see.

As both @zhuyaliang and you are not seeing any regression, I'll go ahead and merge this, and do the support work if need be :)

raveit65:
> 
> No, tabbing behavior was already OK, the goal of this PR is to make Orca able to understand what section labels (the bold text) relate to. With this PR, Orca will use those labels to announce the settings groups, just like you visually see.
> 
I didn't say that. Please read carefully ;)
The tab is now marked as active with Menta theme with a pointing line, which is new.

raveit65:
Is it with Blue-Menta, i guess mate-screenshot is stealing the focus with your screenshot.
mate-window-properties

cwendling:
@raveit65 and you didn't see this before merging this PR??

Using <kbd>Alt+PrintScreen</kbd>:
Before After

raveit65:
Dunno, i think it was new

cwendling:
Anyway, it doesn't really matter I think, and if anybody hates it enough we could try and bring the focus to the first actual control, but well.  We'll see.

PR #705, opened 2023-05-24
windows: Fix spacing in the Compositing Manager tab

cwendling:
Trivial fix for inconsistent spacing in the *Compositing Manager* tab, which was 1px bigger than the other tabs (and 13 is an unusual amount, where 12 is a common one)


zhuyaliang (approved):
6, 12, and 18 are commonly used values for ```Gtk3```

PR #706, opened 2023-06-04
fix typo

faelpinho:


raveit65:
We ar using transifex for translation. Become a member there of your language group.

PR #720, opened 2023-09-11
Update glib2 dependency version from 2.50 to 2.64

zhuyaliang:
Recently, I plan to add a new subcategory function for mate-c-c: display system details. For details, please refer to gnome-c-c ->Details
2023-09-11 17-07-31屏幕截图

https://gitlab.gnome.org/GNOME/gnome-control-center/-/blob/main/panels/info-overview/cc-system-details-window.c

This will use the functions provided in glib-2.64


raveit65:
I am still wondering about your Centos8 screenshot https://centos.pkgs.org/8-stream/centos-baseos-x86_64/glib2-2.56.4-159.el8.x86_64.rpm.htm 
It still use glib2-2.56.4 same as rocky8. But i don't think that i will ever update rel/centos/rocky8 with mate-1.28.
rhel/centos-stream/Rocky9 using [glib2-2.68.4-6.el9.x86_64.rpm](https://rockylinux.pkgs.org/9/rockylinux-baseos-x86_64/glib2-2.68.4-6.el9.x86_64.rpm.html), so i am fine with the change.
As always i do not know the situation in other distros.

raveit65 (approved):
LGTM

lukefromdc:
Debian Unstable and testing are now on 2.78.0, Stable on 2.74.6 Oldstable (bullseye) on 2.66.8 and Oldoldstable (buster) 2.58.3

zhuyaliang:
I need to use ```g_get_os_Info``` requires ```glib-2.64``` or higher
https://docs.gtk.org/glib/func.get_os_info.html

lukefromdc (approved):
I don't see any reason for a problem with this on Debian as only oldoldstable uses anything older than this. They ship MATE 1.20 and presumably that won't ever change. Testing and Unstable already use 1.78.0. A test build went fine, no issues.

While I cannot speak for the Debian devs, this looks fine to me. Approving

lukefromdc:
OK to merge this or does anyone else want to weigh in?

PR #722, opened 2023-09-28
Add new capplets: system info

zhuyaliang:
Other desktops offer this feature, and we should also add it. I use the ```MATE desktop```, but every time I display system information, I have to open the ```gnome-control-center info-overview```

### DIFF
--- a/capplets/system-info/mate-system-info.c
+++ b/capplets/system-info/mate-system-info.c
@@ -1,0 +1,758 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright (C) 2023 MATE Developers
+ * Copyright (C) 2019 Purism SPC
+ * Copyright (C) 2017 Mohammed Sadiq <sadiq@sadiqpk.org>
+ * Copyright (C) 2010 Red Hat, Inc
+ * Copyright (C) 2008 William Jon McCann <jmccann@redhat.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+#include <gtk/gtk.h>
+#include <glibtop/fsusage.h>
+#include <glibtop/mountlist.h>
+#include <glibtop/mem.h>
+#include <glibtop/sysinfo.h>
+#include <udisks/udisks.h>
+#include <sys/utsname.h>
+
+#ifdef GDK_WINDOWING_WAYLAND
+#include <gdk/gdkwayland.h>
+#endif
+#ifdef GDK_WINDOWING_X11
+#include <gdk/gdkx.h>
+#endif
+
+#include "info-cleanup.h"
+#include "mate-system-info.h"
+#include "mate-system-info-resources.h"
+
+struct _MateSystemInfo
+{
+    GtkDialog    parent_instance;
+    GtkWidget   *logo_image;
+    GtkWidget   *hostname_row;
+    GtkListBox  *hardware_box;
+    GtkWidget   *hardware_model_row;
+    GtkWidget   *memory_row;
+    GtkWidget   *processor_row;
+    GtkWidget   *graphics_row;
+    GtkWidget   *disk_row;
+    GtkListBox  *os_box;
+    GtkWidget   *kernel_row;
+    GtkWidget   *virtualization_row;
+    GtkWidget   *windowing_system_row;
+    GtkWidget   *mate_version_row;
+    GtkWidget   *os_name_row;
+    GtkWidget   *os_type_row;
+};
+
+G_DEFINE_TYPE (MateSystemInfo, mate_system_info, GTK_TYPE_DIALOG)
+
+static void
+set_lable_style (GtkWidget  *lable,
+                 const char *color,
+                 int         font_szie,
+                 const char *text,
+                 gboolean    blod)
+{
+    g_autofree gchar *lable_text = NULL;
+
+    if (color == NULL)
+    {
+        lable_text = g_strdup_printf ("<span weight=\'light\'font_desc=\'%d\'><b>%s</b></span>", font_szie, text);
+    }
+    else
+    {
+        if(blod)
+        {
+            lable_text = g_strdup_printf ("<span foreground=\'%s\'weight=\'light\'font_desc=\'%d\'><b>%s</b></span>",
+                             color,
+                             font_szie,
+                             text);
+        }
+        else
+        {
+            lable_text = g_strdup_printf ("<span foreground=\'%s\'weight=\'light\'font_desc=\'%d\'>%s</span>",
+                            color,
+                            font_szie,
+                            text);
+        }
+    }
+
+    gtk_label_set_markup (GTK_LABEL(lable), lable_text);
+}
+
+static void
+mate_system_info_row_fill (GtkWidget  *row,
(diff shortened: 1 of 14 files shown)


lukefromdc (commented):
The Travis issue appears to be nothing more than not specifying udisks2 (in Debian) as required. My build finished with NO build warnings, and ran fine under both x11 and Wayland. Better yet, it even distinguished whether x11 or wayland was the windowing system. All this on the FIRST try!

Two minor issues did come up:
1: this runtime warning appeared 4x at runtime:
`(mate-system-info:421033): Pango-WARNING **: 23:34:55.327: Leftover font scales`

2: all the text on the right side of the window (details for each item identified) was greyed out. Other than that it looked good, I tested it in every theme I have installed

lukefromdc:
Also note that the strings in ` ReplaceStrings ` can quickly become obsolete as new hardware is introduced, which could become a maintainance hassle. Most of this information can be extracted from the output of `hwinfo` though using that as a shell command called from inside the capplet could include some complex parsing. The popopen() call can do it, but this outputs to a FILE variable which then has to be read into a string, which requires knowing how many bytes to allocate as a buffer and so on. There's almost certainly a better to do that, I just don't know it yet.

zhuyaliang:
@lukefromdc```ReplaceStrings``` come from ```gnome```, and we can follow ```gnome``` to update the processor
> all the text on the right side of the window (details for each item identified) was greyed ou

To distinguish between names and data, I used two colors

Will this warning appear under all themes ```(mate-system-info:421033): Pango-WARNING **: 23:34:55.327: Leftover font scales```

raveit65:
> Will this warning appear under all themes? `(mate-system-info:421033): Pango-WARNING **: 23:34:55.327: Leftover font scales`

Yes, i get the warning with all our themes.
Can we use marco windows-decoration please?
Same like the other capplets from Mate.

lukefromdc (commented):
Just retested, we are still getting CSD window decorations instead of the normal marco/compiz SSD decorations.

zhuyaliang:
@lukefromdc @raveit65 Warning removed

raveit65:
@zhuyaliang 
confirmed, the warning is gone.
But the windows-decoration is a csd window which is wrong for Mate.
The window should be a normal dialog window wich use metacyty/marco decorations.

raveit65:
I found a commit where you removed the headerbar in mate-user-admin https://github.com/zhuyaliang/user-admin/commit/137086edb76d88605506020790a5f6e9a19e86fe

capplets/system-info/mate-system-info.c
```diff
@@ -0,0 +1,745 @@
...
+    mate_system_info_set_row (MATE_SYSTEM_INFO (self));
+}
+
+GtkWidget *
+mate_system_info_new (void)
+{
+    GtkWidget *dialog;
+
+    dialog = g_object_new (MATE_TYPE_SYSTEM_INFO,
+                           "use-header-bar", TRUE,
```
raveit65, line 739: Here is use-header-bar!

lukefromdc:
```suggestion
                           "use-header-bar", TRUE,

    GtkWidget *dialog;

    dialog = g_object_new (MATE_TYPE_SYSTEM_INFO,
                           "use-header-bar", TRUE,
```

lukefromdc: I just tested


    dialog = g_object_new (MATE_TYPE_SYSTEM_INFO,
```suggestion
                           "use-header-bar", TRUE,
```
                           "use-header-bar", FALSE,

and as expected that gave us the normal SSD decorations

lukefromdc: dialog = g_object_new (MATE_TYPE_SYSTEM_INFO,
```suggestion
                           "use-header-bar", TRUE,
```
                           "use-header-bar", FALSE,

lukefromdc: I cannot figure out how to get the suggestions to work, but I just tested using
```

    dialog = g_object_new (MATE_TYPE_SYSTEM_INFO,
                           "use-header-bar", FALSE,
```
and that fixed the problem as expected and gave us the normal window decorations for MATE

zhuyaliang: Determine whether to use header bar based on configuration

lukefromdc: Works fine in x11, but creates a problem in wayland, where we do not have an xsettings manager. A lot of dialogs there go to the GTK default, which is CSD. When running under wayland, we should unconditionally set
use_header_bar to FALSE or this dialog will be the only dialog from mate-control-center to get CSD decorations when compositors such as wayfire or sway are set to create server-side decorations. A compositor set to create client side decorations will be no difference of course.

In general, we should not use header bars anywhere in MATE, and should only allow them as a user option when this won't come with a side order of some users being forced to use them. If you want to make this user selectable in x11, we need to ensure we can get SSD decorations in wayland.

raveit65:
With last commit the decoration in X11 is fine.
@zhuyaliang Can we have a minimize button?

Under wayland it doesn't display in the m-c-c window, same like many other from our capplets.
mcc_mate-system-info_wayland
For some reasons i can start it from common tasks and it has a csd decoration.

@lukefromdc 
Those capplets are not visible in mcc-window and menus under wayland.
- system-info
- time-admin
- mate-media
- mate-user-manager
- and some more

@zhuyaliang 
Your recording tool does everything right under wayland. It displays in mcc and it has the correct decoration, ihmo.
mcc_mate-screen_recording_wayland

raveit65:
Last change gives the best windows-decoration experience under wayland, ihmo.

zhuyaliang:
@raveit65 Added minimize status button

raveit65:
minimize status button works fin in X11.

lukefromdc:
All the capplets are displayed in wayland unconditionally if the compositor is invoked wirh dbus-launch

raveit65:
Indeed, with starting the session from konsole with dbus-lunch the capplets are there in menus and mcc.
I am using a display-manager (gdm), better for switching beween user accounts.
Honestly, i don't see any progress with wayland.
How should a normal user use it without a login manager.....
This is only experimental for freaks.

raveit65 (approved):
LGTM

lukefromdc:
> Indeed, with starting the session from konsole with dbus-lunch the capplets are there in menus and mcc. I am using a login-manager (gdm), better for switching beween user accounts. Honestly, i don't see any progress with wayland. How should a normal user use it without a login manager..... This is only experimental for freaks.

Experimental right now. We need wayfire and/or another wlroots compositor to support looking for our .ini file to handle installing a session from a package manager. I have a simple one that works NOW from sddm with a couple scripts, but one of the three files has to be ~/.config/wayfire.ini to work. I will have to submit a feature request to wayfire for looking for session files in /etc

Real point to the wayland work is to be able to continue if distros start dropping x11: insurance

lukefromdc:
Note that I cannot create a session package installable by the session manager w/o work done on one or more of the compositors to recognize a MATE  startup configuration file.

lukefromdc (approved):
Approving as we need to get this done. EDIT: last commit FIXED wayland window decorations, working around the issue that some not all gsettings values are being overrridden in wayland. Thanks!

lukefromdc:
Setting GTK settings on wayland with GTK4 can be done with a portal-but apparently NOT with GTK3-but themes can include all of the necessary settings in their settings.ini file.  I will have to confirm that the xsettings-manager can override them in x11, if that part works each theme can install sane defaults for wayland. Alternately, these files  could go into a separate "mate-wayland-settings" package to keep it out of default xorg installs. That will kill one of the biggest hassles in using Wayland (including the one in this package,  meaning this is ready to merge), leaving only the session issue.

"https://docs.gtk.org/gtk3/class.Settings.html

GtkSettings provide a mechanism to share global settings between applications.

 On the X window system, this sharing is realized by an XSettings manager that is usually part of the desktop environment, along with utilities that let the user change these settings. In the absence of an Xsettings manager, GTK+ reads default values for settings from settings.ini files in /etc/gtk-3.0, $XDG_CONFIG_DIRS/gtk-3.0 and $XDG_CONFIG_HOME/gtk-3.0. These files must be valid key files (see GKeyFile), and have a section called Settings. Themes can also provide default values for settings by installing a settings.ini file next to their gtk.css file."

lukefromdc:
I now have a working, installable mate-wayland session, for now in my own repo at
https://github.com/lukefromdc/mate-wayland-session

PR #725, opened 2023-10-09
font-viewer: Fix compilation warnings and replace deprecated functions

zhuyaliang:


lukefromdc (approved):
Builds and runs fine, font-view.c build with no warnings at all with my normal build options. No new runtime warnings. Note that with glib 2.78 we do get four gfile warnings on opening mate-appearance-properties from the control center but those are shared with master. No additonal warnings on opening anything related to fonts within it.

Code wise this is a local reimplementation of a deprecated GTK3 function, and dealing with these reduces our exposure to bitrot.

raveit65:
@lukefromdc 
Code is from mate-font-viewer application

raveit65 (approved):
LGTM,
deprecation warnings are gone and mate-font-viewer works like before.

PR #727, opened 2023-10-09
appearance: Remove useless match_wm_manager_init function

zhuyaliang:
```appearance``` is no need to call ```match_wm_manager_init``` because we no longer use ```match_wm_manager_get_current``` and ```match_wm_manager_spawn_config_tool_for_current```, This is a useless initialization
Similarly, ```libwindow-settings``` is an outdated and unnecessary API that has not been called by any project

### DIFF
--- a/capplets/appearance/Makefile.am
+++ b/capplets/appearance/Makefile.am
@@ -40,7 +40,6 @@ mate_appearance_properties_SOURCES = \
 AM_CFLAGS = -DMATE_DESKTOP_USE_UNSTABLE_API
 
 mate_appearance_properties_LDADD = \
-	$(top_builddir)/libwindow-settings/libmate-window-settings.la \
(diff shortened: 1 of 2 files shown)


raveit65:
I am still wondering when mate_wm_manager_get_current and mate_wm_manager_spawn_config_tool were used.
Currently we definitely don't use them.

raveit65:
Btw. I guess you meant mate_wm_manager_get_current and mate_wm_manager_spawn_config_tool. Looks like a typo or not?

raveit65 (approved):
LGTM,
mate-appearance-properties works well.

lukefromdc (approved):
No suprises from mate-appearance-properties launched from m-c-c with this applied, and build went clean after reconfiguration. Looks good on this end

PR #729, opened 2023-10-10
typing-break: Fix compilation warnings and replace deprecated functions

zhuyaliang:


lukefromdc (commented):
Got a build failure here with this error when dropping these three files into a prior build:

```
drw-utils.c: In function ‘set_pixmap_background’:
drw-utils.c:126:9: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘gtk_widget_realize’
  126 |         gtk_widget_realize (window);
      |         ^~~~~~~~~~~~~~~~~~
drw-utils.c:172:9: error: ‘cairo_region’ undeclared (first use in this function); did you mean ‘_cairo_region’?
  172 |         cairo_region = cairo_region_create ();
      |         ^~~~~~~~~~~~
      |         _cairo_region
drw-utils.c:172:9: note: each undeclared identifier is reported only once for each function it appears in
make: *** [Makefile:676: mate_typing_monitor-drw-utils.o] Error 1

```

zhuyaliang:
@lukefromdc The problem has been fixed. Please test again

lukefromdc (approved):
Works fine now, typing break behavior same as it is with master. Note that with this OR with master, for the first minute or so after turning typing break on, the icon remains grey creating the false impression this isn't working until what is probably one minute counts down and the icon updates. This is same with master and with this PR, be aware of this when testing.

zhuyaliang:
After the code is merged, I will remove the deprecated  ```gtk_status_icon```  APIS and make the ```indicator``` a mandatory dependency

typing-break/drwright.c
```diff
@@ -834,21 +842,59 @@ DrWright *
...
+	      "</object>"
+	    "</child>"
+	    "<child>"
+	      "<object class=\"GtkMenuItem\" id=\"take_break_item\">"
+	        "<property name=\"visible\">1</property>"
+	        "<property name=\"label\" translatable=\"yes\">_Take a Break</property>"
+	        "<property name=\"use-underline\">1</property>"
+	        "<property name=\"action-name\">win.TakeABreak</property>"
+	      "</object>"
+	    "</child>"
```
cwendling, line 894: What about using `GMenuModel`?  It makes for a cleaner UI definition, and is pretty decent to use for normal cases.
The only thing I don't know how to do is change the label for the *Take a Break* entry as is done in the [AppIndicator code path](https://github.com/mate-desktop/mate-control-center/pull/729/files#diff-bf19f230b4e90cc7db2c43e63f3c383e4ee5f78ffa6e44fab01b75f00929baf2L550), but there might be a solution -- and I kinda wonder why it's doing that, but I haven't tested with appindicator.

cwendling: Maybe something like that (appindicatior thing still not fixed)
```diff
diff --git a/typing-break/drwright.c b/typing-break/drwright.c
index f0ab564b..b47568e2 100644
--- a/typing-break/drwright.c
+++ b/typing-break/drwright.c
@@ -71,7 +71,8 @@ struct _DrWright {
 
 	DrwMonitor      *monitor;
 
-	GtkBuilder      *ui_manager;
+	GtkWidget       *menu;
+	GSimpleAction   *break_action;
 
 	DrwState         state;
 	DrwTimer        *timer;
@@ -551,8 +552,10 @@ update_status (DrWright *dr)
 	}
 
 #ifdef HAVE_APP_INDICATOR
-	item = (GtkWidget *)gtk_builder_get_object (dr->ui_manager, "take_break_item");
+/* FIXME: 
+	item = (GtkWidget *)gtk_builder_get_object (dr->ui_builder, "take_break_item");
 	gtk_menu_item_set_label (GTK_MENU_ITEM (item), str);
+*/
 #else
 	gtk_status_icon_set_tooltip_text (dr->icon, str);
 #endif /* HAVE_APP_INDICATOR */
@@ -585,7 +588,6 @@ gsettings_notify_cb (GSettings *settings,
 		 gpointer     user_data)
 {
 	DrWright  *dr = user_data;
-	GtkWidget *item;
 
 	if (!strcmp (key, "type-time")) {
 		dr->type_time = 60 * g_settings_get_int (settings, key);
@@ -601,8 +603,7 @@ gsettings_notify_cb (GSettings *settings,
 		dr->enabled = g_settings_get_boolean (settings, key);
 		dr->state = STATE_START;
 
-		item = (GtkWidget *)gtk_builder_get_object (dr->ui_manager, "take_break_item");
-		gtk_widget_set_sensitive (item, dr->enabled);
+		g_simple_action_set_enabled (dr->break_action, dr->enabled);
 
 		update_status (dr);
 	}
@@ -628,10 +629,8 @@ popup_preferences_cb (GSimpleAction *action,
 	DrWright *dr = data;
 	GdkScreen *screen;
 	GError    *error = NULL;
-	GtkWidget *menu;
 
-	menu = (GtkWidget *)gtk_builder_get_object (dr->ui_manager, "pop_menu");
-	screen = gtk_widget_get_screen (menu);
+	screen = gtk_widget_get_screen (dr->menu);
 
 	if (!mate_gdk_spawn_command_line_on_screen (screen, "mate-keyboard-properties --typing-break", &error)) {
 		GtkWidget *error_dialog;
@@ -680,17 +679,13 @@ popup_menu_cb (GtkWidget *widget,
 	       guint      activate_time,
 	       DrWright  *dr)
 {
-	GtkWidget *menu;
-
-	menu = (GtkWidget *)gtk_builder_get_object (dr->ui_manager, "pop_menu");
-
-	gtk_menu_popup (GTK_MENU (menu),
+	gtk_menu_popup (GTK_MENU (dr->menu),
 			NULL,
 			NULL,
 			gtk_status_icon_position_menu,
 			dr->icon,
-			0,
-			gtk_get_current_event_time ());
+			button,
+			activate_time);
 }
 #endif /* HAVE_APP_INDICATOR */
 
@@ -772,8 +767,7 @@ init_app_indicator (DrWright *dr)
 					  APP_INDICATOR_STATUS_PASSIVE);
 	}
 
-	indicator_menu = (GtkWidget *)gtk_builder_get_object (dr->ui_manager, "pop_menu");
-	app_indicator_set_menu (dr->indicator, GTK_MENU (indicator_menu));
+	app_indicator_set_menu (dr->indicator, GTK_MENU (dr->menu));
 	app_indicator_set_attention_icon (dr->indicator, TYPING_MONITOR_ATTENTION_ICON);
 
 	update_status (dr);
@@ -842,57 +836,34 @@ DrWright *
 drwright_new (void)
 {
 	DrWright  *dr;
-	GtkWidget *menu;
-	GtkWidget *item;
+	GtkBuilder *ui_builder;
+	GMenuModel *menu_model;
 	GSettings *settings;
-	GActionGroup *action_group;
+	GSimpleActionGroup *action_group;
 
 	static const gchar *ui_description =
-	  "<interface>"
-	    "<object class=\"GtkImage\" id=\"menu_icon_pre\">"
-	      "<property name=\"visible\">True</property>"
-	      "<property name=\"can_focus\">False</property>"
-	      "<property name=\"icon-name\">preferences-desktop</property>"
-	    "</object>"
-	    "<object class=\"GtkImage\" id=\"menu_icon_about\">"
-	      "<property name=\"visible\">True</property>"
-	      "<property name=\"can_focus\">False</property>"
-	      "<property name=\"icon-name\">help-about</property>"
-	    "</object>"
-	    "<object class=\"GtkMenu\" id=\"pop_menu\">"
-	        "<property name=\"visible\">1</property>"
-	    "<child>"
-	      "<object class=\"GtkImageMenuItem\" id=\"preferences_item\">"
-	        "<property name=\"visible\">1</property>"
-	        "<property name=\"label\" translatable=\"yes\">_Preferences</property>"
-	        "<property name=\"use-underline\">1</property>"
-	        "<property name=\"image\">menu_icon_pre</property>"
-	        "<property name=\"action-name\">win.Preferences</property>"
-	      "</object>"
-	    "</child>"
-	    "<child>"
-	      "<object class=\"GtkImageMenuItem\" id=\"about_item\">"
-	        "<property name=\"visible\">1</property>"
-	        "<property name=\"label\" translatable=\"yes\">_About</property>"
-	        "<property name=\"use-underline\">1</property>"
-	        "<property name=\"image\">menu_icon_about</property>"
-	        "<property name=\"action-name\">win.About</property>"
-	      "</object>"
-	    "</child>"
-	    "<child>"
-	      "<object class=\"GtkSeparatorMenuItem\">"
-	        "<property name=\"visible\">1</property>"
-	      "</object>"
-	    "</child>"
-	    "<child>"
-	      "<object class=\"GtkMenuItem\" id=\"take_break_item\">"
-	        "<property name=\"visible\">1</property>"
-	        "<property name=\"label\" translatable=\"yes\">_Take a Break</property>"
-	        "<property name=\"use-underline\">1</property>"
-	        "<property name=\"action-name\">win.TakeABreak</property>"
-	      "</object>"
-	    "</child>"
-	 "</interface>";
+		"<interface>"
+			"<menu id=\"pop_menu\">"
+				"<section>"
+					"<item>"
+						"<attribute name=\"label\" translatable=\"yes\">_Preferences</attribute>"
+						"<attribute name=\"action\">win.Preferences</attribute>"
+						"<attribute name=\"icon\">preferences-desktop</attribute>"
+					"</item>"
+					"<item>"
+						"<attribute name=\"label\" translatable=\"yes\">_About</attribute>"
+						"<attribute name=\"action\">win.About</attribute>"
+						"<attribute name=\"icon\">help-about</attribute>"
+					"</item>"
+				"</section>"
+				"<section>"
+					"<item>"
+						"<attribute name=\"label\" translatable=\"yes\">_Take a Break</attribute>"
+						"<attribute name=\"action\">win.TakeABreak</attribute>"
+					"</item>"
+				"</section>"
+			"</menu>"
+		"</interface>";
 
 	dr = g_new0 (DrWright, 1);
 
@@ -912,22 +883,27 @@ drwright_new (void)
 		setup_debug_values (dr);
 	}
 
-	dr->ui_manager = gtk_builder_new ();
+	action_group = g_simple_action_group_new ();
+	g_action_map_add_action_entries (G_ACTION_MAP (action_group),
+					 action_entries,
+					 G_N_ELEMENTS (action_entries),
+					 dr);
+	dr->break_action = G_SIMPLE_ACTION (g_action_map_lookup_action (G_ACTION_MAP (action_group), "TakeABreak"));
+	g_simple_action_set_enabled (dr->break_action, dr->enabled);
 
-	action_group = (GActionGroup*)g_simple_action_group_new ();
-	gtk_builder_set_translation_domain (dr->ui_manager, GETTEXT_PACKAGE);
+	ui_builder = gtk_builder_new ();
+#ifdef ENABLE_NLS
+	gtk_builder_set_translation_domain (ui_builder, GETTEXT_PACKAGE);
+#endif /* ENABLE_NLS */
+	gtk_builder_add_from_string (ui_builder, ui_description, -1, NULL);
+	menu_model = G_MENU_MODEL (gtk_builder_get_object (ui_builder, "pop_menu"));
 
-	g_action_map_add_action_entries (G_ACTION_MAP (action_group),
-                                        action_entries,
-                                        G_N_ELEMENTS (action_entries),
-                                        dr);
-	gtk_builder_add_from_string (dr->ui_manager, ui_description, -1, NULL);
+	dr->menu = gtk_menu_new_from_model (menu_model);
 
-	menu = (GtkWidget *)gtk_builder_get_object (dr->ui_manager, "pop_menu");
-	gtk_widget_insert_action_group (menu, "win", action_group);
+	g_object_unref (ui_builder);
 
-	item = (GtkWidget *)gtk_builder_get_object (dr->ui_manager, "take_break_item");
-	gtk_widget_set_sensitive (item, dr->enabled);
+	gtk_widget_insert_action_group (dr->menu, "win", G_ACTION_GROUP (action_group));
+	g_object_unref (action_group);
 
 	dr->timer = drw_timer_new ();
 	dr->idle_timer = drw_timer_new ();
```

typing-break/drw-utils.c
```diff
@@ -168,13 +169,15 @@ set_pixmap_background (GtkWidget *window)
...
+	gdc = gdk_window_begin_draw_frame (gtk_widget_get_window (window), cairo_region);
+	cr = gdk_drawing_context_get_cairo_context (gdc);
 	gdk_cairo_set_source_pixbuf (cr, tmp_pixbuf, 0, 0);
 	cairo_paint (cr);
 
 	g_object_unref (tmp_pixbuf);
-
-	cairo_destroy (cr);
+	gdk_window_end_draw_frame (gtk_widget_get_window (window), gdc);
+	cairo_region_destroy (cairo_region);
```
cwendling, line 180: this looks fairly suspicious, but I guess it might work on x11.  I can't test, because that doesn't seem to work neither on master or on this branch though, and I get a black break screen (with or without compositing enabled).  If you know a way to make that work, I could try again.

typing-break/drwright.c
zhuyaliang: @cwendling This will result in the loss of icons, which is unacceptable for icon enthusiasts

zhuyaliang: I have made some modifications to the code, but I have decided to continue using ```GtkMenu```

typing-break/drw-utils.c
zhuyaliang: In ```X11```, ```gdk_screen_is_composed``` returned ```FALSE```, and ```set_pixmap_background``` was not called, so I am unable to test.I replaced it according to the gtk3 help manual

https://docs.gtk.org/gdk3/func.cairo_create.html

raveit65: It works fine for me on current master using compositor and x11, i started it from status-applet.
The UI is black with white/red text in the middle of the monitor.
Graphic card issue?
I am using AMD integrated gpu.
... i will test current state of the branch.

typing-break/drwright.c
```diff
@@ -607,23 +606,25 @@ gsettings_notify_cb (GSettings *settings,
...
 		dr->state = STATE_BREAK_SETUP;
 		maybe_change_state (dr);
 	}
 }
 
 static void
-popup_preferences_cb (GtkAction *action, DrWright *dr)
+popup_preferences_cb (GSimpleAction *action,
+		              GVariant      *parameter,
+					  gpointer       data)
```
raveit65, line 621: Indents.
```suggestion
popup_preferences_cb (GSimpleAction *action,
	              GVariant      *parameter,
	              gpointer       data)
```

raveit65:
The typing-break UI looks and works like before, same for notification-area applet.

lukefromdc:
What is the status of this?

typing-break/drwright.c
cwendling: > This will result in the loss of icons, which is unacceptable for icon enthusiasts

No it would not, GMenu supports icons just fine (see the `<attribute name="icon">help-about</attribute>` for example), and I tested it working.

> I have made some modifications to the code, but I have decided to continue using `GtkMenu`

Fair enough.  I just find that it's a lot of XML compared to the previous version whereas `GMenu` tries to be a little more concise in describing menus, but it's not necessarily a problem.  And admittedly I don't know how to solve the appindicator case as mentioned, so it's more than fair.

typing-break/drw-utils.c
cwendling: @raveit65 yeah I have the same black background, but some what I read it apparently is supposed to draw a pattern over the desktop, not a black window… anyhow, if black is OK, it worked for me as well.
Using i915 on an IntelHD integrated GPU from a Clarkdale-era CPU.

typing-break/drwright.c
```diff
@@ -866,18 +901,22 @@ drwright_new (void)
...
+	ui_builder = gtk_builder_new ();
 #ifdef ENABLE_NLS
-	gtk_action_group_set_translation_domain (action_group, GETTEXT_PACKAGE);
+	gtk_builder_set_translation_domain (ui_builder, GETTEXT_PACKAGE);
 #endif /* ENABLE_NLS */
-	gtk_action_group_add_actions (action_group, actions, G_N_ELEMENTS (actions), dr);
-	gtk_ui_manager_insert_action_group (dr->ui_manager, action_group, 0);
-	gtk_ui_manager_add_ui_from_string (dr->ui_manager, ui_description, -1, NULL);
+	gtk_builder_add_from_string (ui_builder, ui_description, -1, NULL);
+	dr->menu = (GtkWidget *)gtk_builder_get_object (ui_builder, "pop_menu");
```
cwendling, line 909: You need to hold a reference to the menu, otherwise destroying the `GtkBuilder` below will also destroy the menu as it's not tied to anything. Currently the menu doesn't work for me, and I get:
```
(mate-typing-monitor:1578107): GLib-GObject-WARNING **: 11:53:54.367: invalid unclassed pointer in cast to 'GtkMenu'

(mate-typing-monitor:1578107): Gtk-CRITICAL **: 11:53:54.367: gtk_menu_popup: assertion 'GTK_IS_MENU (menu)' failed
```
when trying to open the status icon's menu.


```suggestion
	dr->menu = (GtkWidget *) g_object_ref (gtk_builder_get_object (ui_builder, "pop_menu"));
```

cwendling (changes requested):

raveit65:
@cwendling 
UI with a so called "black screen" and a view elements.
typing-break

cwendling:
@raveit65 yeah it's what I see as well; but again I don't think it's what the code is supposed to do.  Anyway, it doesn't looks like a regression from master, so it doesn't matter for this PR.

typing-break/drw-utils.c
cwendling: This will probably work on X11 anyway, and IIUC nobody sees a difference with master, so I think it's OK to use that.

cwendling (approved):
LGTM, and seems to work as before.
I did **not** test AppIndicator support.

raveit65:
@cwendling 
Fedora use this default since you implemented it.
```
gsettings get org.mate.panel enable-sni-support
true
```
I guess AppIndicator support works when i see the menu of the applet, or not?

cwendling:
@raveit65 if the applet is built with it, yes.

raveit65 (approved):
LGTM

lukefromdc:
Indicator (SNI) support for this capplet still looks for the
older Ubuntu libappindicator not the current ayatana version. Porting might just be a matter of allowing the capplet to build against the ayatana version.

raveit65:
Yes, mcc use `--enable-appindicator` build flag.

PR #730, opened 2023-10-10
capplets/display: Fix compilation warnings and replace deprecated functions

zhuyaliang:


lukefromdc (approved):
With default CFLAGS the capplet builds with no warnings at all with this applied. Runtime behavior includeing colors to mark monitors (one monitor tested as I am away from home) were the same as before. Even the color indicating my monitor in the monitor name popup window, both of the squares one inside the other within the display dialog, and the monitor name in the status icon popup menu looked the same as before

raveit65 (approved):
LGTM
- fixes compiler warnings about deprecated code
- code change seems valid
- no new build warnings
- no runtime warnings
- mate-display-properties runs fine

PR #731, opened 2023-10-11
Remove  "libwindow settings" that are no longer in use

zhuyaliang:
```libwindow-settings``` is no longer used by any warehouse or code. We should remove it


lukefromdc (approved):
windows capplet still works with this removed, same behavior both in marco and with (unsupported dialog appears) compiz

raveit65 (approved):
LGTM

PR #732, opened 2023-10-18
typing-break: Force dependency on appindicator3-0.1 and remove deprecated GtkStatusIcon

zhuyaliang:
```
gsettings set org.mate.typing-break enabled true
```


typing-break/drwright.c
```diff
@@ -30,9 +30,7 @@
 #include <gtk/gtk.h>
 #include <gio/gio.h>
 
-#ifdef HAVE_APP_INDICATOR
 #include <libappindicator/app-indicator.h>
```
cwendling, line 33: The ayatana-appindicator include is `libayatana-appindicator/app-indicator`

cwendling (changes requested):
Doesn't work yet with ayatana-appindicator (and probably not with appindicator either though), but I guess the cleanup is cool.

Here is a proposed patch to cleanup and fix ayatana-appindicator support: https://github.com/mate-desktop/mate-control-center/commit/2103e55ee1c5b26d9d81606615be07f68483f9e0 (branch of the same name as this PR on my fork if you want to fetch it)

typing-break/main.c
```diff
@@ -98,26 +77,6 @@ main (int argc, char *argv[])
 		return 0;
 	}
 
-#ifndef HAVE_APP_INDICATOR
-	if (!no_check && !have_tray ()) {
```
cwendling, line 102: should we have something similar for appindicator?  It should be doable checking whether there's a status notifier host registered.  Not saying it's needed, just a thought.

cwendling (commented):
Seems to work from very basic testing with Ayatana AppIndicator

typing-break/main.c
zhuyaliang: done

lukefromdc (commented):
Builds fine and solves the issue with an "empty" grey hatched icon showing up when first setting a test work interval of one or two minutes. Instead of being delayed by one minute, the  green time lapse bar icon showed up immediately.

However, the icon does not update, instead staying a full height green bar right up until the screen locks. I do not normally use this feature but recall the icon's green bar dropping with time and flashing in the last minute or so before the "take a break" screen lock renders.

Postponing breaks does not work and that button is greyed out but that's been true in master for a while

Tested with ayatana indicators, a 4K monitor, and window-scaling = 2, icon renders clear and sharp.

typing-break/main.c
cwendling: Do you think checking for a tray makes sense?  For us it does because the notification area does both SN host and system tray (on x11, on Wyaland there won't be a system tray I guess), but a better check would be whether there's the right DBus thingy at the right place -- not that I can tell you how to do that, but some inspiration from the SN host in notification-area might possibly help.

Then again, I'm not entirely sure how important that warning is in real life.

zhuyaliang: I plan to use Dbus to check if the notification area exists, but I have encountered a problem，I'm not sure if it's my usage issue
```
dbus-send --session --print-reply=literal --dest=org.kde.StatusNotifierWatcher  /StatusNotifierWatcher org.freedesktop.DBus.Introspectable.Introspect
```
https://gitlab.gnome.org/GNOME/glib/-/issues/3150

cwendling: I barely know anything about DBus, let alone GDbus, but this seems to work (indeed `g_dbus_proxy_new_for_bus_sync()` don't seem to want to fail, but it's not necessarily a problem here):

```c
#include <stdio.h>
#include <glib.h>
#include <gio/gio.h>

static gboolean
is_status_notifier_host_available (void)
{
    const gchar *service = "org.kde.StatusNotifierWatcher";
    const gchar *path = "/StatusNotifierWatcher";
    const gchar *interface = "org.kde.StatusNotifierWatcher";
    const gchar *property = "IsStatusNotifierHostRegistered";

    g_autoptr (GDBusProxy) proxy = NULL;
    g_autoptr (GError) error = NULL;

    proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SESSION,
                                           G_DBUS_PROXY_FLAGS_DO_NOT_CONNECT_SIGNALS |
                                           G_DBUS_PROXY_FLAGS_DO_NOT_AUTO_START,
                                           NULL, service, path, interface,
                                           NULL, &error);

    if (proxy == NULL || error)
    {
        g_warning ("g_dbus_proxy_new_for_bus_sync() failed: %s", error->message);
        return FALSE;
    }

    g_autoptr (GVariant) variant = g_dbus_proxy_get_cached_property (proxy, property);
    if (! variant)
    {
        g_warning ("Property %s.%s not found!", interface, property);
        return FALSE;
    }
    if (! g_variant_is_of_type (variant, G_VARIANT_TYPE_BOOLEAN))
    {
        g_warning ("Property %s.%s is invalid", interface, property);
        return FALSE;
    }
    
    return g_variant_get_boolean(variant);
}

int main(void)
{
    fprintf (stderr, "Status Notifier host available? %s\n",
             is_status_notifier_host_available () ? "yes" : "no");

    return 0;
}
```

Checking `IsStatusNotifierHostRegistered` property sounds to me like a good thing to do, as I'd think it not only says there's a watcher, but also an actual host.
Anyway, take this with a grain of salt, again I'm not very knowledgeable about this stuff.

zhuyaliang: work well

lukefromdc:
Latest force-pushed worked same as my last test, icon does not change

lukefromdc (approved):
Icon updating isn't working properly here on the status icon with master either. In fact, the status icon seems to show the countdown backwards: the grey "empty" bar-png icon
renders FIRST, and bar-green.png gradually fills it from the bottom up-instead of emptying it from the top down. Flashing red for last few seconds works for the status icon only.
Everything else works, and the menu catches the actual time remaining. 

If there is an easy fix for the icon behavior we can do it here (maybe we are not getting gdbus working properly here?), if not we can leave that for a later PR as the indicator behavior (a constant unchanging icon) isn't as counterintuitive as what the status icon is doing is.

zhuyaliang:
@lukefromdc Leave this issue to future PR for resolution

lukefromdc:
No problem, I've approved this PR and am fine with merging it. We'll deal with the icon updating later

typing-break/main.c
```diff
@@ -29,28 +29,44 @@
...
+	if (proxy == NULL || error)
+	{
+		return FALSE;
+	}
+	g_autoptr (GVariant) variant = g_dbus_proxy_get_cached_property (proxy, NOTIFIERAREA_PROPERTY);
+	if (! variant)
+	{
 		return FALSE;
 	}
+	if (! g_variant_is_of_type (variant, G_VARIANT_TYPE_BOOLEAN))
```
raveit65, line 63: Build warning in CI  (fedora with normal compiler flags.
https://app.travis-ci.com/github/mate-desktop/mate-control-center/jobs/611885988#L6843
```
main.c:63:39: warning: Casting a non-structure type to a structure type and accessing a field can lead to memory access errors or data corruption [alpha.core.CastToStruct]

        if (! g_variant_is_of_type (variant, G_VARIANT_TYPE_BOOLEAN))

                                             ^~~~~~~~~~~~~~~~~~~~~~

/usr/include/glib-2.0/glib/gvarianttype.h:50:46: note: expanded from macro 'G_VARIANT_TYPE_BOOLEAN'

#define G_VARIANT_TYPE_BOOLEAN              ((const GVariantType *) "b")

                                             ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

raveit65:
Beside from the build warning applet works same as before in fedora with UBUNTU_APPINDICATOR support.

typing-break/main.c
zhuyaliang: done

raveit65 (approved):
LGTM

raveit65:
@cwendling 
We have unresolved conversations?

typing-break/main.c
zhuyaliang: @cwendling Can I merge? I will delete libslab for the next PR and need to complete it on this basis

lukefromdc:
@cwendling, is this ready to go or should we wait?

typing-break/main.c
cwendling: IMO this should be restored.  Yes, `GVariantType` is kind of weird, but it's actually just strings.  I guess the use of a virtual structure is to have a dedicated type and do some type validation, or room for future expansion maybe.
At any rate, the code was correct and not actually dangerous.  If newer (?) compilers produce a warning, it might be interesting to report that to upstream GLib to see if they can come up with a clever trick to hide the spurious ones, or whatever they can do to silence this.

If we don't check the variant type, and it's not actually a boolean, the following `g_variant_get_boolean()` might warn, give unexpected results, or potentially crash in the future.  The current implementation look pretty safe, but [the documentation](https://docs.gtk.org/glib/method.Variant.get_boolean.html) is clear that it's not supported.

cwendling (commented):
Apart from possibly the inline comment, LGTM

raveit65:
If we really should produce build warnings this should be commented in code.
Otherwise this can be removed in future again by mistake because we forgot why we accept a warning.
Ihmo, i would prefer to leave code warnings free and wait for bugreport if it really produce a crash.

cwendling:
> If we really should produce build warnings this should be commented in code. Otherwise this can be removed in future again by mistake because we forgot why we accept a warning.

Fair enough.  But it's actually *none of our own code*, it's just using GLib API as intended… so we could add a comment about that API being broken in that regard, but that'd look a tad odd :)

> Ihmo, i would prefer to leave code warnings free and wait for bugreport if it really produce a crash.

Fair enough.  But I'd rather not have crashes in the first place, because that's way nastier than a warning.  Either way, in practice this probably doesn't actually matter because the property is gonna be of the right type, and currently `g_variant_get_boolean()` is safe enough.

So just as you guys like best, I don't mind.

typing-break/main.c
zhuyaliang: It's right to be careful， but ```g_variant_get_boolean()``` is secure enough, and the ```IsStatusNotifierHostRegistered```  type is known and controllable, so ```g_variant_get_boolean()```  will not fail,type check can be removed

raveit65:
Good, than let us merge PR like it is :)

PR #737, opened 2023-10-25
Remove libslab library

zhuyaliang:
libslab is only used in mate-c-c and there is no need to provide a library


raveit65 (approved):
LGTM

lukefromdc (approved):
Builds and runs fine here, note that building (splitting out) libslab was listed as a build option as long as I can remember. If nothing but the control center uses it, putting part of the control center into a shared library is just needless complexity.

lukefromdc:
Does anyone else want to review this or should be merge it?

PR #739, opened 2023-12-29
Populate in wayland

lukefromdc:
*When running under Wayland, show all of the capplets that are useful in wayland now or which are intended to be in the future

*Show the control center in the main menu when running under wayland. Note that we cannot simply use `XDG_CURRENT_DESKTOP=MATE` or we block users from setting certain GNOME gsettings values differently for x11 and wayland.  We use `XDG_CURRENT_DESKTOP=MATE-wayland` to distinguish the sessions but the control center's .desktop files need to accomodate this.

*This also keeps most x11-only capplets out of the control center when running under wayland


lukefromdc:
Rebased after https://github.com/mate-desktop/mate-control-center/pull/740

lukefromdc:
Closing as it is no longer needed after
https://github.com/mate-desktop/mate-wayland-session/commit/3a262255213695a34e35c69ce0711d91d5d7a632
and using "MATE-wayland" for an OnlyShowIn string violates the .desktop file specifications barring a new registration of a new string

PR #740, opened 2023-12-29
Meson: fix mate-time-admin directory error

lukefromdc:
*Ensure we attempt to load files from `mate-time-admin/map/ `not ` mate-time-admin/map`

*Otherwise mate-time-admin built with meson cannot find the files and errors out, crashing on an attempt to load (with /usr/share/  as  data directory`)/usr/share/mate-time-admin/mapbackward `instead of 
`/usr/share/mate-time-admin/map/backward`


zhuyaliang (approved):
Confirm fixing the issue

Issue #744, opened 2024-02-23
[Clarification and Documentation] systemd requirement of mate-control-center?


rubyFeedback:
Hey guys,

This is not to get into any debate about pro or con; I am mostly fine by decisions made either way, but perhaps
a bit of context, in addition to the issue I report here.

I am currently compiling all of mate and I have almost finished. Example of versions I have:

    Reporting the installed versions of the mate-desktop components:

      Mate-Desktop:         1.28.1
      Mate-Utils:           1.28.0
      Mate-Common:          1.28.0
      Mate-Keyboard:        1.28.0
      Mate-Media:           1.28.0
      Mate-Mixer:           1.28.0
      Mate-Applets:         1.28.0
      Mate-Weather:         1.28.0
      Mate-Control-Center:  1.26.1
      Mate-Panel:           1.28.0
      Mate-Power-Manager:   1.28.0
      Marco:                marco: symbol lookup error: marco: undefined symbol: meta_verbose_real
      Mate-Screensaver:     1.28.0
      Atril:                1.28.0
      Eom:                  1.28.0
      Pluma:                1.28.0
      Engrampa:             1.28.0
      Caja:                 1.28.0
      Caja-extensions:      1.28.0

Don't worry about the marco error - that is an unrelated error. I'll fix that lateron.

So I am now trying to update Mate-Control-Center:  1.26.1 to its 1.28.x variant.

When I run GNU configure, though, it tells me this:

    checking for polkit-gobject-1... yes
    checking for mate-desktop-2.0 >= 1.27.1... yes
    checking for libxml-2.0... yes
    checking for libmarco-private >= 1.17.0... yes
    checking for gmodule-2.0 glib-2.0 >= 2.64.0 gio-2.0 gtk+-3.0 >= 3.22.0... yes
    checking for gio-2.0... yes
    checking for libgtop-2.0... yes
    checking for udisks2... yes
    checking for systemd >= 248... no
    configure: error: Package requirements (systemd >= 248) were not met:

    No package 'systemd' found

So systemd is now required. Note that mate-control-center is the only package that
has this issue; at the least I don't recall any of the other mate 1.28.x components
have this problem. (I do not use systemd, instead I use the LFS scripts udev variants.)

I tried:

    --disable-systemd

But this does not exist so I assume systemd is now **mandatory** for mate-control-center.

This is a bit unusual though. How much of the code in mate-control-center depends on
[...]

PR #748, opened 2024-03-27
Add time-admin dependency check on systemd

zhuyaliang:
time-admin depends on systemd. If systemd is not used in the systemd, time-admin is disabled

### DIFF
--- a/capplets/Makefile.am
+++ b/capplets/Makefile.am
@@ -8,10 +8,13 @@ SUBDIRS = \
 	keybindings \
 	mouse \
 	network \
-	time-admin \
 	windows \
 	about-me \
 	system-info
+
+if HAVE_SYSTEMD
+SUBDIRS += time-admin
+endif
 
 DIST_SUBDIRS = \
 	common \


zhuyaliang:
@cwendling I am not very familiar with autoconf, can I modify it this way?

lukefromdc:
Looks to me like we are setting a variable, then setting it to a non-null value only if systemd is detected. I also don't know the build systems that well but this should have the same effect when systemd is not present as the 
` = \`
lines as that's what it should expand to in those cases

We also support meson for mate-control-center, though only one actually HAS to work for non-systemd distros to build m-c-c today, the point to any meson work is an eventual transition to it. That is important especially if autotools ever goes unmaintained.

zhuyaliang:
After this PR is accepted, I will add ```meson``` support for checking ````systemd```

capplets/Makefile.am
```diff
@@ -22,7 +29,7 @@ DIST_SUBDIRS = \
 	keyboard \
 	mouse \
 	network \
-	time-admin \
+	$(TIMEADMIN) \
```
cwendling, line 32: You must not define it conditionally in `DIST_SUBDIRS`, so that tarballs & co are properly created.  Just remove this change, having it conditional in `SUBDIRS` is enough to do what you want.  You can [check the documentation](https://www.gnu.org/software/automake/manual/automake.html#Conditional-Subdirectories) which is surprisingly clear :)

Actually, `DIST_SUBDIRS` doesn't really seem useful at all, but let's not touch this in this PR as it's unrelated.

capplets/Makefile.am
```diff
@@ -1,3 +1,10 @@
+if HAVE_SYSTEMD
+TIMEADMIN = time-admin
+else
+TIMEADMIN =
```
cwendling, line 4: There's no real need to define `TIMEADMIN` empty in the false clause (undefined MAKE variables expand to an empty value), but I guess it doesn't hurt either.

cwendling (changes requested):
Almost OK :)
You can also verify it does the right thing if `make dist` creates a tarball with the time-admin capplet sources when run from a build tree configured with systemd disabled.

capplets/Makefile.am
zhuyaliang: I used a more concise method

cwendling (approved):
LGTM

zhuyaliang:
Fix #744

PR #749, opened 2024-03-28
appearance in wayland: Support changing theme, icon theme, and font

lukefromdc:
*If we are in a wayland session, write the gtk theme and icon theme to both org.mate and org.gnome *This changes wayland and xwayland themes simultaniously *startup with x11 backend, if we are under wayland without xwayland we don't have mate-settings-daemon anyway

### DIFF
--- a/capplets/appearance/appearance-font.c
+++ b/capplets/appearance/appearance-font.c
@@ -733,6 +733,7 @@ cb_show_details (GtkWidget *button,
 void font_init(AppearanceData* data)
 {
 	GtkWidget* widget;
+	int ret;
 
 	data->font_details = NULL;
 	data->font_groups = NULL;
@@ -780,6 +781,40 @@ void font_init(AppearanceData* data)
 			  "changed::" WINDOW_TITLE_USES_SYSTEM_KEY,
 			  G_CALLBACK (marco_changed),
 			  data);
+
+	/*In a wayland session we must manage MATE font settings for xwayland
+	 *and also manage GNOME font settings for native wayland applications
+	 *
+	 *First find out if we are running under a wayland session
+	 */
+
+	ret = system ("killall -0 -e Xwayland");
+	/*If we are, set the GNOME interface keys too*/
+	if (ret == 0)
+	{
+		widget = appearance_capplet_get_widget(data, "application_font");
+		g_settings_bind (data->interface_gnome_settings,
+				 GTK_FONT_KEY,
+				 G_OBJECT (widget),
+				 "font-name",
+				 G_SETTINGS_BIND_DEFAULT);
+
+		widget = appearance_capplet_get_widget (data, "document_font");
+		g_settings_bind (data->interface_gnome_settings,
+				 DOCUMENT_FONT_KEY,
+				 G_OBJECT (widget),
+				 "font-name",
+				 G_SETTINGS_BIND_DEFAULT);
+
+/* The monospace font seems to apply properly if and only if set only for MATE
+		widget = appearance_capplet_get_widget (data, "monospace_font");
+		g_settings_bind (data->interface_gnome_settings,
+				 MONOSPACE_FONT_KEY,
+				 G_OBJECT (widget),
+				 "font-name",
+				 G_SETTINGS_BIND_DEFAULT);
+*/
+	}
 
 	g_signal_connect (appearance_capplet_get_widget (data, "add_new_font"), "clicked", G_CALLBACK (cb_add_new_font), data);
 
--- a/capplets/appearance/appearance-font.c
+++ b/capplets/appearance/appearance-font.c
@@ -753,7 +753,6 @@ cb_show_details (GtkWidget *button,
 void font_init(AppearanceData* data)
 {
 	GtkWidget* widget;
-	int ret;
 
 	data->font_details = NULL;
 	data->font_groups = NULL;
@@ -804,13 +803,9 @@ void font_init(AppearanceData* data)
 
 	/*In a wayland session we must manage MATE font settings for xwayland
 	 *and also manage GNOME font settings for native wayland applications
-	 *
-	 *First find out if we are running under a wayland session
+	 *so if not running under x11, set the GNOME interface keys too
 	 */
-
-	ret = system ("killall -0 -e Xwayland");
-	/*If we are, set the GNOME interface keys too*/
-	if (ret == 0)
+	if (!(GDK_IS_X11_DISPLAY ((gdk_display_get_default()))))
 	{
 		widget = appearance_capplet_get_widget(data, "application_font");
 		g_settings_bind (data->interface_gnome_settings,
--- a/capplets/appearance/appearance-font.c
+++ b/capplets/appearance/appearance-font.c
@@ -804,8 +804,10 @@ void font_init(AppearanceData* data)
 	/*In a wayland session we must manage MATE font settings for xwayland
 	 *and also manage GNOME font settings for native wayland applications
 	 *so if not running under x11, set the GNOME interface keys too
+	 *Ignore this if for any reason the GNOME schema was not found,
+	 *As we can only set this for compositors using either the MATE or the GNOME gsettings
 	 */
-	if (!(GDK_IS_X11_DISPLAY ((gdk_display_get_default()))))
+	if (data->interface_gnome_settings)
 	{
 		widget = appearance_capplet_get_widget(data, "application_font");
 		g_settings_bind (data->interface_gnome_settings,
--- a/capplets/appearance/appearance-font.c
+++ b/capplets/appearance/appearance-font.c
@@ -421,20 +421,40 @@ dpi_from_pixels_and_mm (int pixels, int mm)
 }
 
 static double
-get_dpi_from_x_server (void)
+get_dpi_from_x_server_or_monitor (void)
 {
   GdkScreen  *screen;
+  GdkDisplay *display;
   double dpi;
 
   screen = gdk_screen_get_default ();
+  display = gdk_display_get_default();
 
   if (screen) {
     double width_dpi, height_dpi;
 
-    Screen *xscreen = gdk_x11_screen_get_xscreen (screen);
+    if (GDK_IS_X11_DISPLAY (display))
+    {
+      Screen *xscreen = gdk_x11_screen_get_xscreen (screen);
 
-    width_dpi = dpi_from_pixels_and_mm (WidthOfScreen (xscreen), WidthMMOfScreen (xscreen));
-    height_dpi = dpi_from_pixels_and_mm (HeightOfScreen (xscreen), HeightMMOfScreen (xscreen));
+      width_dpi = dpi_from_pixels_and_mm (WidthOfScreen (xscreen), WidthMMOfScreen (xscreen));
+      height_dpi = dpi_from_pixels_and_mm (HeightOfScreen (xscreen), HeightMMOfScreen (xscreen));
+    }
+    else
+    {
+      GdkMonitor *monitor;
+      GdkRectangle geometry = {0};
+
+      /*FIXME: we have to use the main monitor for this
+       *which may not always be the leftmost monitor
+       *Separate per-monitor settings for this would be ideal
+       */
+      monitor = gdk_display_get_monitor (display, 0);
+      gdk_monitor_get_geometry (monitor, &geometry);
+
+      width_dpi = dpi_from_pixels_and_mm (geometry.width, gdk_monitor_get_width_mm(monitor));
+      height_dpi = dpi_from_pixels_and_mm (geometry.height, gdk_monitor_get_height_mm(monitor));
+    }
 
     if (width_dpi < DPI_LOW_REASONABLE_VALUE || width_dpi > DPI_HIGH_REASONABLE_VALUE ||
         height_dpi < DPI_LOW_REASONABLE_VALUE || height_dpi > DPI_HIGH_REASONABLE_VALUE)
@@ -465,7 +485,7 @@ dpi_load (GSettings     *settings,
   dpi = g_settings_get_double (settings, FONT_DPI_KEY);
 
   if (dpi == 0)
-    dpi = get_dpi_from_x_server ();
+    dpi = get_dpi_from_x_server_or_monitor ();
 
   dpi *= (double)scale;
   dpi = CLAMP(dpi, DPI_LOW_REASONABLE_VALUE, DPI_HIGH_REASONABLE_VALUE);
--- a/capplets/appearance/appearance-main.c
+++ b/capplets/appearance/appearance-main.c
@@ -31,6 +31,7 @@
 #include "theme-thumbnail.h"
 #include "activate-settings-daemon.h"
 #include "capplet-util.h"
+#include <gdk/gdkx.h>
 
 static AppearanceData *
 init_appearance_data (int *argc, char ***argv, GOptionContext *context)
@@ -56,8 +57,12 @@ init_appearance_data (int *argc, char ***argv, GOptionContext *context)
 
   data->filechooser_settings = g_settings_new (FILECHOOSER_SCHEMA);
   data->interface_settings = g_settings_new (INTERFACE_SCHEMA);
-  /*Loading this unconditionally may be cheaper than another system() call*/
-  data->interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
+
+  if (!(GDK_IS_X11_DISPLAY (gdk_display_get_default())))
+    data->interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
+  else
+    data->interface_gnome_settings = NULL;
+
   data->marco_settings = g_settings_new (MARCO_SCHEMA);
   data->mouse_settings = g_settings_new (MOUSE_SCHEMA);
   data->font_settings = g_settings_new (FONT_RENDER_SCHEMA);
--- a/capplets/appearance/appearance-main.c
+++ b/capplets/appearance/appearance-main.c
@@ -38,6 +38,9 @@ init_appearance_data (int *argc, char ***argv, GOptionContext *context)
   AppearanceData *data = NULL;
   GtkBuilder *ui;
 
+  /*Set the backend to use x11 (must use xwayland in wayland session*/
+  gdk_set_allowed_backends ("x11"); 
+
   theme_thumbnail_factory_init (*argc, *argv);
   capplet_init (context, argc, argv);
   activate_settings_daemon ();
@@ -56,6 +59,8 @@ init_appearance_data (int *argc, char ***argv, GOptionContext *context)
 
   data->filechooser_settings = g_settings_new (FILECHOOSER_SCHEMA);
   data->interface_settings = g_settings_new (INTERFACE_SCHEMA);
+  /*Loading this unconditionally may be cheaper than another system() call*/
+  data->interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
   data->marco_settings = g_settings_new (MARCO_SCHEMA);
   data->mouse_settings = g_settings_new (MOUSE_SCHEMA);
   data->font_settings = g_settings_new (FONT_RENDER_SCHEMA);
@@ -162,10 +167,6 @@ main (int argc, char **argv)
       	N_("[WALLPAPER...]") },
       { NULL, 0, 0, G_OPTION_ARG_NONE, NULL, NULL, NULL }
     };
-
-
-  /*Set the backend to use x11 (must use xwayland in wayland session*/
-  gdk_set_allowed_backends ("x11");
 
   option_context = g_option_context_new (NULL);
   g_option_context_add_main_entries (option_context, option_entries, GETTEXT_PACKAGE);
--- a/capplets/appearance/appearance-main.c
+++ b/capplets/appearance/appearance-main.c
@@ -163,6 +163,10 @@ main (int argc, char **argv)
       { NULL, 0, 0, G_OPTION_ARG_NONE, NULL, NULL, NULL }
     };
 
+
+  /*Set the backend to use x11 (must use xwayland in wayland session*/
+  gdk_set_allowed_backends ("x11");
+
   option_context = g_option_context_new (NULL);
   g_option_context_add_main_entries (option_context, option_entries, GETTEXT_PACKAGE);
 
--- a/capplets/appearance/appearance-main.c
+++ b/capplets/appearance/appearance-main.c
@@ -58,11 +58,20 @@ init_appearance_data (int *argc, char ***argv, GOptionContext *context)
   data->filechooser_settings = g_settings_new (FILECHOOSER_SCHEMA);
   data->interface_settings = g_settings_new (INTERFACE_SCHEMA);
 
+  data->interface_gnome_settings = NULL;
+  /*Load the gnome interface schema if we are running under wayland and it is present*/
   if (!(GDK_IS_X11_DISPLAY (gdk_display_get_default())))
-    data->interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
-  else
-    data->interface_gnome_settings = NULL;
+  {
+    GSettingsSchemaSource *source = g_settings_schema_source_get_default ();
 
+    if (source)
+    {
+      GSettingsSchema *schema = g_settings_schema_source_lookup (source, INTERFACE_GNOME_SCHEMA, TRUE);
+
+      if (schema)
+        data->interface_gnome_settings = g_settings_new_full (schema, NULL, NULL);
+    }
+  }
   data->marco_settings = g_settings_new (MARCO_SCHEMA);
   data->mouse_settings = g_settings_new (MOUSE_SCHEMA);
   data->font_settings = g_settings_new (FONT_RENDER_SCHEMA);
--- a/capplets/appearance/appearance-main.c
+++ b/capplets/appearance/appearance-main.c
@@ -38,9 +38,6 @@ init_appearance_data (int *argc, char ***argv, GOptionContext *context)
   AppearanceData *data = NULL;
   GtkBuilder *ui;
 
-  /*Set the backend to use x11 (must use xwayland in wayland session*/
-  gdk_set_allowed_backends ("x11"); 
-
   theme_thumbnail_factory_init (*argc, *argv);
   capplet_init (context, argc, argv);
   activate_settings_daemon ();
--- a/capplets/appearance/appearance.h
+++ b/capplets/appearance/appearance.h
@@ -58,6 +58,7 @@
 #define GTK_THEME_KEY                "gtk-theme"
 #define ICON_THEME_KEY               "icon-theme"
 #define INTERFACE_SCHEMA             "org.mate.interface"
+#define INTERFACE_GNOME_SCHEMA       "org.gnome.desktop.interface"
 #define MENU_ICONS_KEY               "menus-have-icons"
 #define MONOSPACE_FONT_KEY           "monospace-font-name"
 #define TOOLBAR_STYLE_KEY            "toolbar-style"
@@ -92,6 +93,10 @@ typedef struct {
 	GSettings* caja_settings;
 	GSettings* filechooser_settings;
 	GSettings* interface_settings;
+	/*We have to accomodate wayland theme setting here
+	 *whether we are using it or not
+	 */
+	GSettings* interface_gnome_settings;
 	GSettings* marco_settings;
 	GSettings* mouse_settings;
 	GSettings* font_settings;
--- a/capplets/common/mate-theme-apply.c
+++ b/capplets/common/mate-theme-apply.c
@@ -56,18 +56,14 @@ mate_meta_theme_set (MateThemeMetaInfo *meta_theme_info)
   GSettings *notification_settings = NULL;
   gchar *old_key;
   gint old_key_int;
-  gboolean wayland;
 
-  /*Find out whether we are running under wayland or x11 */
-
-  if (GDK_IS_X11_DISPLAY ((gdk_display_get_default())))
-    wayland = FALSE;
-
+  /*We only want to touch GNOME settings under Wayland */
+  if (GDK_IS_X11_DISPLAY (gdk_display_get_default()))
+    interface_gnome_settings = NULL;
   else
-    wayland = TRUE;
+    interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
 
   interface_settings = g_settings_new (INTERFACE_SCHEMA);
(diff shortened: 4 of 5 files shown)


capplets/common/mate-theme-apply.c
```diff
@@ -48,13 +49,22 @@ void
...
   GSettings *notification_settings = NULL;
   gchar *old_key;
   gint old_key_int;
+  int ret;
+
+  /*Find out if we are running under wayland and have xwayland
+   *Sending signal 0 with killall doesn't kill the process but
+   *the return value tells us whether the process exists
+   */
+  ret = system ("killall -0 -e Xwayland");
```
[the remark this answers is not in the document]
lukefromdc: I just tested that change, and unfortunately I could not get it to work. The reason is that with system() we get the return value of the program run, and that's what we are using to distinguish the wayland session from an x11 session while running under Xwayland.

With `g_spawn_command_line_async() `we don't. Since finding nothing on running `killall -0 -e <anything>` does not seem to be considered an error, using GError doesn't seem to give a way of distinguishing this either. While this function is a gboolean returning TRUE or FALSE, it seems to return TRUE unconditionally when running 
`g_spawn_command_line_async("killall -0 -e Xwayland", NULL);`
I tested that by comparing behavior in both wayland and x11 sessions

A different issue cropped up trying to use` g_spawn_command_line_async()` to start mate-appearance-properties from caja in https://github.com/mate-desktop/caja/pull/1771
 if using mate-control-center 1.28.0 or current git master, we cannot open mate-appearance-properties without passing `GDK_BACKEND=x11` as an environmental variable. We can get by without it entirely in caja if this PR is applied to mate-control-center, but that would cause caja to have different behavior in wayland depending on which version of m-c-c is installed, shortly after a major release. Not a hard dependency due to the fallback, but something that might generate bug reports.

 `g_spawn_command_line_async()` does not seem to be able to pass environmental variables, and neither do the similar commands such as those that underlie `caja_launch_application_from_command`. That one can be made to work by setting it to open a command in terminal, but then we have that showing on the desktop plus all the code to run a totally unneeded terminal. Three things worked there:  `g_spawn_command_line_async()` or `caja_launch_application_from_command()` with this PR applied to m-c-c and not trying to pass any environmental variables, or `caja_launch_application_from_command()` setting terminal usabe to TRUE, which allowed  `GDK_BACKEND=x11` to be passed and the background page of this capplet to open.

I agree that it would be good to avoid `system()` as it is heavy, but this seems to be hard to avoid here. In this case, an alternative mode of detecting that we are in a wayland session from an app running in Xwayland might be able get us around using `system()` which is rather heavy for this job.  In the caja case by comparison, the fork etc should be less of a concern as mate-appearance-properties being launched from there is a whole new toplevel program anyway. There too though we have to run two checks, there too I tested g_spawn_command_line_async() and the checks didn't work.

Thoughts?

lukefromdc:
Latest force-push adds control of cursor theme, note however that cursor theme setting in wayland has been reported to be buggy no matter how it's done.

capplets/common/mate-theme-apply.c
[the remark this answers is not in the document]
lukefromdc: Since https://github.com/mate-desktop/mate-control-center/pull/749/commits/3835297e5add6c2d3327d5813b861939e41d1713 the appearance capplet can now run under native wayland. That means we no longer need that system call to detect the session so
https://github.com/mate-desktop/mate-control-center/pull/749/commits/4832386262c8fbadf6cce5bacb0b571b15b99338

removes it, and this should make lighter code

lukefromdc:
https://github.com/mate-desktop/caja/pull/1771 updated to take advantage of this without depending on it.

Still a draft as I am hoping to get theme customization to apply in wayland. we have fonts, theme, and icon theme now but customization of a theme is only applying on closing and reopening m-c-c.

lukefromdc:
We also need to suppress the "settings manager is not running" error dialog in wayland, since the settings manager is used only to control Xwayland. Alternately the dialog could be modified under wayland to indicate specifically that settings won't be applied in Xwayland, and suppress it when the appearance capplet is opened to the background page

lukefromdc:
Since a caja PR takes advantage of this, customized theme support can be included with this if I can find what is needed to apply customization in this timeframe, or it can wait for a followup.

This  gives a lot of new functionality in a wayland session as matters stand now,partially resolving one of the biggest user hassles: how to set the theme and font.

capplets/appearance/appearance-font.c
```diff
@@ -421,20 +421,40 @@ dpi_from_pixels_and_mm (int pixels, int mm)
...
+    else
+    {
+      GdkMonitor *monitor;
+      GdkRectangle geometry = {0};
+
+      /*FIXME: we have to use the main monitor for this
+       *which may not always be the leftmost monitor
+       *Separate per-monitor settings for this would be ideal
+       */
+      monitor = gdk_display_get_monitor (display, 0);
```
zhuyaliang, line 452: Why is it ```0``` here? If multiple monitors are connected, will there be a problem here? I'm not sure

capplets/common/mate-theme-apply.c
```diff
@@ -48,13 +50,26 @@ void
...
   gint old_key_int;
+  gboolean wayland;
+
+  /*Find out if we are running under wayland and have xwayland
+   *Sending signal 0 with killall doesn't kill the process but
+   *the return value tells us whether the process exists
+   */
+  if (GDK_IS_X11_DISPLAY ((gdk_display_get_default())))
+    wayland = FALSE;
+
```
zhuyaliang, line 67: Extra spaces

capplets/appearance/appearance-font.c
lukefromdc: We are setting the font here, and right now I have yet to figure out how to do this on a per-monitor basis. First priority on that would be showing the caja desktop on all monitors. Monitor 0 (leftmost or only monitor on single monitor setup) is always present, and in the wayland session it's where caja's desktop goes unless manually started from a terminal on another monitor.

capplets/common/mate-theme-apply.c
lukefromdc: This file was converted some time ago to GNOME style 2 space indents. That is the indent under the `if` statement,  same under the `else` statement. If this makes the file hard to read, I can use brackets instead if you or anyone else prefers.

capplets/appearance/appearance-font.c
zhuyaliang: Would it be better to use ```gdk_display_get_primary_monitor``` ?

lukefromdc:
Not sure it would alwaya work

https://docs.gtk.org/gdk3/method.Display.get_primary_monitor.html

Says this can return NULL when no primary monitor is configured, and in Wayland the concept of a primary monitor seems to have not been implemented.

capplets/appearance/appearance-font.c
zhuyaliang: Would it be better to use  ```gdk_display_get_monitor_at_window``` instead

lukefromdc:
That might work, will try it when I get home

capplets/appearance/appearance-font.c
lukefromdc: We don't have the window the capplet is using within `get_dpi_from_x_server_or_monitor`, and we don't have it in the calling function either. Thus we'd have to pass the window through a lot of other functions to get it here it seems.

In an ideal multimonitor case, an argument could be made for using the monitor the capplet is on, or for using the monitor the desktop is on. Setting a different font for each monitor would require a lot of other changes, notable additional gsettings values to store additional values in, we'd have to show each possible monitor in the capplet at well

zhuyaliang: I personally think there's no problem. Let others take a look too

zhuyaliang (approved):
Not validated in the Wayland environment, there are no issues with the code logic

lukefromdc:
Hoping we can get a review by another who has the wayland session set up

lukefromdc:
Is anyone else able to test this in a wayland session, should I merge it, or something else?

capplets/appearance/appearance-font.c
```diff
@@ -781,6 +802,40 @@ void font_init(AppearanceData* data)
 			  G_CALLBACK (marco_changed),
 			  data);
 
+	/*In a wayland session we must manage MATE font settings for xwayland
+	 *and also manage GNOME font settings for native wayland applications
+	 *
+	 *First find out if we are running under a wayland session
+	 */
+
+	ret = system ("killall -0 -e Xwayland");
```
cwendling, line 811: That's a bit… meh.  Can't we just check whether the app's display is a Wayland one?  The only probable issue is if the app is run under Xwayland, but maybe it's only fair that it might not control the Wayland session in that case?

capplets/appearance/appearance-font.c
```diff
@@ -781,6 +802,40 @@ void font_init(AppearanceData* data)
...
+	/*If we are, set the GNOME interface keys too*/
+	if (ret == 0)
+	{
+		widget = appearance_capplet_get_widget(data, "application_font");
+		g_settings_bind (data->interface_gnome_settings,
+				 GTK_FONT_KEY,
+				 G_OBJECT (widget),
+				 "font-name",
+				 G_SETTINGS_BIND_DEFAULT);
```
cwendling, line 820: how does this interact if one of the settings change (either the GNOME or MATE one) from outside?  does it sync both, or does this goes crazy?  I would think it would just sync both but better safe than sorry

capplets/appearance/appearance.h
```diff
@@ -58,6 +58,7 @@
 #define GTK_THEME_KEY                "gtk-theme"
 #define ICON_THEME_KEY               "icon-theme"
 #define INTERFACE_SCHEMA             "org.mate.interface"
+#define INTERFACE_GNOME_SCHEMA       "org.gnome.desktop.interface"
```
cwendling, line 61: does this add a strong dependency on GNOME desktop schemas?

capplets/common/mate-theme-apply.c
```diff
@@ -48,13 +50,26 @@ void
...
   gchar *old_key;
   gint old_key_int;
+  gboolean wayland;
+
+  /*Find out if we are running under wayland and have xwayland
+   *Sending signal 0 with killall doesn't kill the process but
+   *the return value tells us whether the process exists
+   */
+  if (GDK_IS_X11_DISPLAY ((gdk_display_get_default())))
+    wayland = FALSE;
```
cwendling, line 66: The comment seems out of sync.
But I like the check better than the system call :)

capplets/appearance/appearance-font.c
cwendling: BTW if it works, maybe we could just bind the two regardless of Wayland?  Does it cause any issue on X11?  It's just that MCC would partially control the GNOME settings as well, but given some apps use those maybe it's only fair?

cwendling: (BTW IMO we should at some point have MSD work under Wayland, but I get that it's very far from trivial, so maybe it's not realistic, at least not for now…)

capplets/appearance/appearance.h
lukefromdc: Not a new one, pretty sure gtk3 itself depends on that, correct me if I am wrong.

capplets/common/mate-theme-apply.c
lukefromdc: Thanks for catching this, looks like two code blocks I wrote at two different times used two different methods

capplets/appearance/appearance-font.c
lukefromdc: All the code does here is write to gsettings. I just checked, having this and dconf-editor open at the same time. If you change it in either one it is applied and immediately reflected in the other.

Binding the GNOME settings when under x11 would probably get complaints from users and even distros, as no doubt some want to have both sessions installed and with different settings. This is no problem under x11 but is impossible under wayland at this time unless we maintain our own compositor (probably a fork of wayfire but with really very few changes). Probably we should submit a PR for wayfire to allow use of MATE or other gsettings for these values

capplets/common/mate-theme-apply.c
lukefromdc: That is fixed, I missed it in the previous commit that removed the same system call elsewhere

lukefromdc:
Last force-push just updates a comment to reflect that in native wayland we no longer have to resort to the system call

capplets/appearance/appearance-font.c
lukefromdc: Also note that we DO in fact need to run mate-settings-daemon in the x11 session or we don't apply nondefault theme/font/icon theme values in Xwayland apps. It controls only settings of apps running under x11 or xwayland, but without it we can no longer sync the theme etc between a wayland session and apps running under xwayland. Therefore, we run m-s-d under the current wayland session for that purpose

capplets/appearance/appearance-font.c
lukefromdc: The original reason for this was that at first I had this whole capplet running under xwayland. Once I got past that and got it running under native wayland, there was no more need for this roundabout approach. One of today's updates removed the leftover instance of this,

capplets/appearance/appearance-font.c
cwendling: > Probably we should submit a PR for wayfire to allow use of MATE or other gsettings for these values.

Yeah I don't know how wayfire works yet, but possibly a plugin similar to Compiz's MATE integration?  Anyway yes if it's at this level that it has to be modified it'd make sense to do so there.  But I guess that can be delayed if you prefer.

capplets/appearance/appearance.h
cwendling: Apparently it does use it for the Wayland backend indeed.  Though, if I read [the code](https://gitlab.gnome.org/GNOME/gtk/-/tree/gtk-3-24/gdk/wayland/gdkscreen-wayland.c#L714) properly it's a soft-dependency, and a fallback one at that, and it uses the DBus `org.freedesktop.portal.Settings` interface on the `org.freedesktop.portal.Desktop` session bus (plus something for the fontconfig settings).  So either that's what Wayfire provides, or we'd just have to provide that instead of an XSettings backend in MSD.

cwendling: So what I mean is that this *does* make MCC depend on the GNOME desktop schemas.  Whether another component needed for the Wayland session already does or not is yet to be determined.

capplets/appearance/appearance-main.c
```diff
@@ -56,6 +56,8 @@ init_appearance_data (int *argc, char ***argv, GOptionContext *context)
 
   data->filechooser_settings = g_settings_new (FILECHOOSER_SCHEMA);
   data->interface_settings = g_settings_new (INTERFACE_SCHEMA);
+  /*Loading this unconditionally may be cheaper than another system() call*/
+  data->interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
```
cwendling, line 60: you could remove the comment or guard with a `! GDK_IS_X11_DISDPLAY()` call

cwendling: BTW if you wanted this schema to be optional (in case it makes any sense to have it optional), you could do something like this:
```c
  GSettingsSchemaSource *source = g_settings_schema_source_get_default ();

  if (source)
  {
    GSettingsSchema *schema = g_settings_schema_source_lookup (source, INTERFACE_GNOME_SCHEMA, TRUE);

    if (schema)
      data->interface_gnome_settings = g_settings_new_full (schema, NULL, NULL);
  }
```

Again, not saying you should do that, as there's no point if we have to have this schema for things to work.

capplets/common/mate-theme-apply.c
```diff
@@ -48,13 +50,24 @@ void
...
   gint old_key_int;
+  gboolean wayland;
+
+  /*Find out whether we are running under wayland or x11 */
+
+  if (GDK_IS_X11_DISPLAY ((gdk_display_get_default())))
+    wayland = FALSE;
+
+  else
+    wayland = TRUE;
```
cwendling, line 67: What about doing something like this instead?
```c
    if (GDK_IS_X11_DISPLAY ((gdk_display_get_default())))
        interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
    else
        interface_gnome_settings = NULL;

/* … */

    if (interface_gnome_settings)
        g_settings_set_string (interface_gnome_settings, 
                               GTK_THEME_KEY, meta_theme_info->gtk_theme_name);
```
Just to avoid coupling `wayland` and `interface_gnome_settings`.

Just an idea, feel free to scrap it :)

capplets/appearance/appearance-main.c
lukefromdc: Since we no longer use system() to tell if we are under wayland or x11, we no longer have to deal unconditionally with that schema. Wayfire btw does indeed depend on it as wayfire (and probably a lot of other non-GNOME/non-KDE compositors) use the GNOME gsettings values.

cwendling: OK so no point in making things more complicated than needed

capplets/common/mate-theme-apply.c
lukefromdc: Just applied this with a new commit, fixing one reversal of the logic as we need the GNOME settings under wayland not under x11. Less loaded into RAM when running under x11 if we never load a schema or create a gsettings object that will never be used

capplets/appearance/appearance-main.c
lukefromdc: Also just confirmed that at least Debian's GTK3 packages directly depend on gsettings-desktop-schemas, which is the package containing the GNOME interface schema

lukefromdc:
Wayfire itself does NOT depend on gsettings-desktop-schemas(debian packaging names here), not sure how it handles settings themes etc (maybe using Adwaita unconditionally?) when this is not installed.

Anyway, I can add checking for the existance of the GNOME interface schema so this will not crash if some distro is shipping GTK without it and assuming that works. Debian's GTK 3 builds at least used to depend on it, I've never tested anything without that package as it would break so much of my system

capplets/appearance/appearance-main.c
cwendling: Hum, [I don't see that](https://packages.debian.org/bookworm/libgtk-3-0), where do you see it?  Not saying it can be installed without, but it would be an indirect dependency)

capplets/appearance/appearance-font.c
```diff
@@ -781,6 +801,38 @@ void font_init(AppearanceData* data)
 			  G_CALLBACK (marco_changed),
 			  data);
 
+	/*In a wayland session we must manage MATE font settings for xwayland
+	 *and also manage GNOME font settings for native wayland applications
+	 *so if not running under x11, set the GNOME interface keys too
+	 *Ignore this if for any reason the GNOME schema was not found,
+	 *As we can only set this for compositors using either the MATE or the GNOME gsettings
+	 */
+	if ((!(GDK_IS_X11_DISPLAY ((gdk_display_get_default())))) && (data->interface_gnome_settings))
```
cwendling, line 810: Do we need both checks if `data->interface_gnome_settings` is only available under X11 (or when we ant to set it)?

cwendling: Also, there's a lot of parentheses, you could at least drop a pair around `gdk_display_get_default()` :wink:

capplets/appearance/appearance-main.c
```diff
@@ -56,6 +57,12 @@ init_appearance_data (int *argc, char ***argv, GOptionContext *context)
 
   data->filechooser_settings = g_settings_new (FILECHOOSER_SCHEMA);
   data->interface_settings = g_settings_new (INTERFACE_SCHEMA);
+
+  if (GDK_IS_X11_DISPLAY ((gdk_display_get_default())))
+    data->interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
+  else
+    data->interface_gnome_settings = NULL;
```
cwendling, line 64: This looks like it is the wrong way around, isn't it?

capplets/common/mate-theme-apply.c
```diff
@@ -68,6 +85,11 @@ mate_meta_theme_set (MateThemeMetaInfo *meta_theme_info)
   if (compare (old_key, meta_theme_info->gtk_theme_name))
     {
       g_settings_set_string (interface_settings, GTK_THEME_KEY, meta_theme_info->gtk_theme_name);
+
+      if ((wayland == TRUE) && (interface_gnome_settings))
```
cwendling, line 89: again, do we really need to check both here?

cwendling:
> Wayfire's Debian package itself does NOT depend on gsettings-desktop-schemas(debian packaging names here), not sure how it handles settings themes etc (maybe using Adwaita unconditionally?) when this is not installed.

I don't know either but as said it seems that on the GTK side a specific DBus name can provide the data, so they could also potentially get the info from some place else (be it a conffile or whatever) and present it that way.  No idea.  And the Debian package could be missing a dep as well if nobody noticed, this happens even in Debian :slightly_smiling_face: 

> Anyway, I can add checking for the existance of the GNOME interface schema so this will not crash if some distro is shipping GTK without it and assuming that works. Debian's GTK 3 builds at least used to depend on it, I've never tested anything without that package as it would break so much of my system

I don't know if it's actually worth it if it looks unlikely anybody is gonna have such a setup, so you can do whichever you prefer I think.  And if you don't do it *and* somebody happen to find a setup where it's a problem, we can always add it…
OTOH if we depend on it we might want to make that explicit. :shrug:

capplets/appearance/appearance-font.c
lukefromdc: Probably not, was just playing cautious. Again, we load that when NOT user x11 though

capplets/appearance/appearance-main.c
lukefromdc: Probably got confused by seeing it shown here reversed

capplets/appearance/appearance-font.c
lukefromdc: Anyway, just took that out in the latest force-push

capplets/appearance/appearance-main.c
lukefromdc: Fixed this in the latest force-push

capplets/common/mate-theme-apply.c
lukefromdc: Changed that to just checking for `wayland = TRUE` in the latest force-push

capplets/common/mate-theme-apply.c
```diff
@@ -68,6 +85,11 @@ mate_meta_theme_set (MateThemeMetaInfo *meta_theme_info)
   if (compare (old_key, meta_theme_info->gtk_theme_name))
     {
       g_settings_set_string (interface_settings, GTK_THEME_KEY, meta_theme_info->gtk_theme_name);
+
+      if (wayland == TRUE)
```
cwendling, line 89: Actually I'd have done it the *other* way around and only checked whether the gsettings was there, assuming whatever the check was for setting it was right:
```suggestion
      if (interface_gnome_settings)
```

capplets/common/mate-theme-apply.c
```diff
@@ -97,12 +119,17 @@ mate_meta_theme_set (MateThemeMetaInfo *meta_theme_info)
...
   g_settings_set_string (marco_settings, MARCO_THEME_KEY, meta_theme_info->marco_theme_name);
+  /*To also control decoration theme in wayland we need a decorator that uses marco themes
+   *and queries gsettings to determine which theme to use to this is a TODO */
 
   /* set the icon theme */
   old_key = g_settings_get_string (interface_settings, ICON_THEME_KEY);
   if (compare (old_key, meta_theme_info->icon_theme_name))
     {
       g_settings_set_string (interface_settings, ICON_THEME_KEY, meta_theme_info->icon_theme_name);
+      if (wayland == TRUE)
```
cwendling, line 130:
```suggestion
      if (interface_gnome_settings)
```

capplets/common/mate-theme-apply.c
```diff
@@ -125,6 +152,12 @@ mate_meta_theme_set (MateThemeMetaInfo *meta_theme_info)
   if (compare (old_key, meta_theme_info->cursor_theme_name))
     {
       g_settings_set_string (mouse_settings, CURSOR_THEME_KEY, meta_theme_info->cursor_theme_name);
+      if (wayland == TRUE)
```
cwendling, line 155:
```suggestion
      if (interface_gnome_settings)
```

capplets/common/mate-theme-apply.c
```diff
@@ -48,11 +50,26 @@ void
 mate_meta_theme_set (MateThemeMetaInfo *meta_theme_info)
 {
   GSettings *interface_settings;
+  GSettings *interface_gnome_settings;
   GSettings *marco_settings;
   GSettings *mouse_settings;
   GSettings *notification_settings = NULL;
   gchar *old_key;
   gint old_key_int;
+  gboolean wayland;
```
cwendling, line 59:
```suggestion
```

capplets/common/mate-theme-apply.c
```diff
@@ -48,11 +50,26 @@ void
...
+  if (GDK_IS_X11_DISPLAY ((gdk_display_get_default())))
+  {
+    wayland = FALSE;
+    interface_gnome_settings = NULL;
+  }
+  else
+  {
+    wayland = TRUE;
+    interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
+  }
```
cwendling, line 72:
```suggestion

  /*We only want to tough GNOME settings under Wayland */
  if (GDK_IS_X11_DISPLAY (gdk_display_get_default()))
    interface_gnome_settings = NULL;
  else
    interface_gnome_settings = g_settings_new (INTERFACE_GNOME_SCHEMA);
```

capplets/appearance/appearance-font.c
```diff
@@ -781,6 +801,38 @@ void font_init(AppearanceData* data)
 			  G_CALLBACK (marco_changed),
 			  data);
 
+	/*In a wayland session we must manage MATE font settings for xwayland
+	 *and also manage GNOME font settings for native wayland applications
+	 *so if not running under x11, set the GNOME interface keys too
+	 *Ignore this if for any reason the GNOME schema was not found,
+	 *As we can only set this for compositors using either the MATE or the GNOME gsettings
+	 */
+	if (!(GDK_IS_X11_DISPLAY (gdk_display_get_default())))
```
cwendling, line 810:
```suggestion
	if (data->interface_gnome_settings)
```

capplets/common/mate-theme-apply.c
lukefromdc: Interesting, that simplifies the code a bit, will do

capplets/common/mate-theme-apply.c
lukefromdc: Latest force-pushed used this simplification, wayland test shows it working

capplets/appearance/appearance-font.c
lukefromdc: included in latest force-push, thanks

capplets/common/mate-theme-apply.c
lukefromdc: included in latest force-push, thanks

capplets/common/mate-theme-apply.c
lukefromdc: done

capplets/common/mate-theme-apply.c
lukefromdc: done

lukefromdc:
Latest changes simplify the logic for detecting x11 vs wayland, looking for x11 once, loading the gnome schema if we are not under x11, then looking for the settings object it is loaded into. That also paves the way to not depending on the schema at all, just not setting the GNOME preferences (thus not theming anything running under wayfire at least) rather than say, crashing.

capplets/appearance/appearance.h
lukefromdc: I will be adding the suggested code to look for the schema and ignore it (and wayland theming) if it is not present

lukefromdc:
last force-push is just a rebase against master

capplets/appearance/appearance-main.c
lukefromdc: https://packages.debian.org/bookworm/libgtk-3-0

lukefromdc: Anyway, latest commit ensures the absence of the GNOME schemas should not crash the control center even when running under wayland, we just don't get theming of apps under wayland that follow the GNOME gsettings values. We could pop up a dialog to warn the user in such cases though

lukefromdc:
Above commit allows installation when the GNOME schemas are not present, we just lose the ability to actually apply the chosen theme to non-xwayland apps when running under wayland with the GNOME schemas absent. No idea if that will ever be used of course

capplets/appearance/appearance-font.c
lukefromdc: Caught those extra  parentheses myself, they were leftovers from the double check I think. Anyway, they are gone, thanks for catching it too.

capplets/appearance/appearance.h
lukefromdc: Does defining this in the header file create a hard dependency when it is not present even when we explicitly check for it before using it, or not?  If it does, we should probably move this out of the header file and just use the schema name directly.

lukefromdc:
Thanks, we can keep that definition than

capplets/appearance/appearance-main.c
cwendling: > https://packages.debian.org/bookworm/libgtk-3-0

I don't see the dep here?  Anyway, as you now only use them if they are present the question is moot.

capplets/common/mate-theme-apply.c
```diff
@@ -48,12 +50,28 @@ void
...
 {
   GSettings *interface_settings;
+  GSettings *interface_gnome_settings;
   GSettings *marco_settings;
   GSettings *mouse_settings;
   GSettings *notification_settings = NULL;
   gchar *old_key;
   gint old_key_int;
 
+  interface_gnome_settings = NULL;
```
cwendling, line 60: you could move this initialization at the declaration right above just like `notification_settings`.  Or keep it here, whichever you prefer.

cwendling:
OK, I didn't test this yet, but code looks OK to me :+1:

lukefromdc:
Last commit just moves initializating of `interface_gnome_settings` to where it is declared, thus removing one line of code.

cwendling:
Looks good :+1:

lukefromdc:
Squashed and merged from the command line with https://github.com/mate-desktop/mate-control-center/commit/1e8c08e51ea38d5faec5f352ea5d00c565a0a63d to allow editing the commit message

lukefromdc:
Thanks everyone for reviewing!
We also have the related caja PR
https://github.com/mate-desktop/caja/pull/1771
which allows setting caja's background from the control center in wayland whenever mate-settings-daemon (as needed for xwayland apps) is running, falls back to the standalone wayland background dialog if it is not running or not installed

Issue #755, opened 2024-06-03
Icon theme is not updated in org.gnome.desktop.interface.icon-theme key after the wayland-compatibility patches landed.


mark-herbert42:
The point is that gtk3 for wayland has a hadcode to read the config exclusively from org.gnome.desktop.interface ignoring all other settings, so the patch series were merged to duplicate some mate,intrface settings into gnome.desktop.interface.  The theme and font settings are duplicated as a charm - but when the icon theme changed only mate dconf is updated.

lukefromdc:
Is THAT why editing a theme has been such a problem? I have been trying to find that, it means there is another place where mate gsettings values are updated that I missed.

mark-herbert42:
yes  - seems that the case. I can not save the new theme with the changed icon set. I can apply old ones - but trying to change the theme in wayand keeps the icons and cursor unchanged. I can change it in dconf editor - no problem, it works. Bit the control center only change the org.mate keys. So far I am lost and have no idea what to do.

lukefromdc:
I need to find the other place that gsettings key is being written to, and open a PR to fix this.

lukefromdc:
Found it: try
https://github.com/mate-desktop/mate-control-center/pull/757 
which fixes this in my testing

mark-herbert42:
757 works fine.  But some new findings

monospaces-font-name   (only this one - other 3 fonts are updated ok)
cursor-theme
cursor-size 

not updated in gnome schema.

lukefromdc:
OK, I can add those as well in the same place. I didn't touch the cursor theming
simply because it has been buggy for me since before the GNOME 2>MATE 
transition. Simple enough to add these though

mark-herbert42:
maybe it is buggy - but manually adding it in dconf does work.  I had issues with it using Xorg/compiz bcz compiz has its own cursor settings and pointers were  switching between compiz settings and mate settings depending over which window the pointer was hoovering. But in wayfire it seems that cursor is strictly obeying org.gnome key.

lukefromdc:
Latest force-push at https://github.com/mate-desktop/mate-control-center/compare/d610d7bbfbe49bc7357026372d838f1afe2c68e2..b29a2e863965cea9f161221835b652758c95f80b
should allow you to set the cursor theme etc. I think I have it hardcoded somewhere since a decade ago to kill an unwanted default that was forced by some kind of bug of that era, thus on my system I cannot test the cursor changes.

[...]

PR #757, opened 2024-06-05
appearance under wayland: fix customizing themes

lukefromdc:
Since wayland compositors such as wayfire follow the GNOME interface gsettings values, we need to set them here too.

*Otherwise we can only reliably select existing metathemes or sometimes a saved custom theme.

### DIFF
--- a/capplets/appearance/appearance-style.c
+++ b/capplets/appearance/appearance-style.c
@@ -30,6 +30,7 @@
 #include "theme-thumbnail.h"
 #include "capplet-util.h"
 #include "appearance-style.h"
+#include <gdk/gdkx.h>
 
 #define GSETTINGS_SETTINGS "GSETTINGS_SETTINGS"
 #define GSETTINGS_KEY      "GSETTINGS_KEY"
@@ -152,6 +153,30 @@ treeview_selection_changed_callback (GtkTreeSelection *selection, guint data)
 
     if (list_value) {
       g_settings_set_string (settings, key, list_value);
+
+    /*Load the gnome interface schema if we are running under wayland and it is present*/
+      if (!(GDK_IS_X11_DISPLAY (gdk_display_get_default())))
+      {
+        GSettingsSchemaSource *source = g_settings_schema_source_get_default ();
+
+        if (source)
+        {
+          GSettingsSchema *schema = g_settings_schema_source_lookup (source, INTERFACE_GNOME_SCHEMA, TRUE);
+          {
+            if (schema)
+            {
+              GSettings *interface_gnome_settings = NULL;
+              interface_gnome_settings = g_settings_new_full (schema, NULL, NULL);
+
+              if ((strcmp (key, GTK_THEME_KEY) == 0) ||
+                 (strcmp (key, ICON_THEME_KEY) == 0) ||
+                 (strcmp (key, CURSOR_THEME_KEY) == 0))
+                      g_settings_set_string (interface_gnome_settings, key, list_value);
+
+            }
+          }
+        }
+      }
     }
   }
 }
--- a/capplets/appearance/appearance-style.c
+++ b/capplets/appearance/appearance-style.c
@@ -482,6 +482,27 @@ static void
 cursor_size_changed_cb (int size, AppearanceData *data)
 {
   g_settings_set_int (data->mouse_settings, CURSOR_SIZE_KEY, size);
+
+  /*If running under wayland and the GNOME interface schema is installed
+   *set the cursor size under the GNOME gsettings too
+   */
+  if (!(GDK_IS_X11_DISPLAY (gdk_display_get_default())))
+  {
+    GSettingsSchemaSource *source = g_settings_schema_source_get_default ();
+
+    if (source)
+    {
+      GSettingsSchema *schema = g_settings_schema_source_lookup (source, INTERFACE_GNOME_SCHEMA, TRUE);
+      {
+        if (schema)
+        {
+          GSettings *interface_gnome_settings = NULL;
+          interface_gnome_settings = g_settings_new_full (schema, NULL, NULL);
+          g_settings_set_int (interface_gnome_settings, CURSOR_SIZE_KEY, size);
+        }
+      }
+    }
+  }
 }
 
 static void


lukefromdc:
Fixes
https://github.com/mate-desktop/mate-control-center/issues/755

lukefromdc:
We got the same Travis issue, again affecting only Fedora. Debian build finished fine
```
 rvm $(travis_internal_ruby) --fuzzy do ruby -S gem install dpl --pre

ERROR:  Error installing dpl:

	The last version of dpl (>= 0) to support your Ruby & RubyGems was 2.0.3.beta.4. Try installing it with `gem install dpl -v 2.0.3.beta.4`

	dpl requires Ruby version >= 3. The current ruby version is 2.7.8.225.

Successfully installed mime-0.4.4

Successfully installed json_pure-2.7.2

Successfully installed excon-0.110.0

Successfully installed travis-packagecloud-ruby-1.1.0

Successfully installed regstry-1.0.15

Successfully installed travis-cl-1.2.4

The command "rvm $(travis_internal_ruby) --fuzzy do ruby -S gem install dpl --pre" failed and exited with 1 during .

Your build has been stopped.
```

capplets/appearance/appearance-style.c
```diff
@@ -152,6 +153,34 @@ treeview_selection_changed_callback (GtkTreeSelection *selection, guint data)
...
+        if (source)
+        {
+          GSettingsSchema *schema = g_settings_schema_source_lookup (source, INTERFACE_GNOME_SCHEMA, TRUE);
+          {
+            if (schema)
+            {
+              GSettings *interface_gnome_settings = NULL;
+              interface_gnome_settings = g_settings_new_full (schema, NULL, NULL);
+
+              g_message("%s", key);
```
cwendling, line 171: This should probably be removed in a production scenario

capplets/appearance/appearance-style.c
```diff
@@ -152,6 +153,34 @@ treeview_selection_changed_callback (GtkTreeSelection *selection, guint data)
...
+
+              if (strcmp (key, ICON_THEME_KEY) == 0)
+                g_settings_set_string (interface_gnome_settings,
+                               ICON_THEME_KEY, list_value);
+
+              if (strcmp (key, GTK_THEME_KEY) == 0)
+                g_settings_set_string (interface_gnome_settings,
+                               GTK_THEME_KEY, list_value);
```
cwendling, line 179:
```suggestion
              if (strcmp (key, ICON_THEME_KEY) == 0 ||
                  strcmp (key, GTK_THEME_KEY) == 0)
                g_settings_set_string (interface_gnome_settings, key, list_value);
```

mbkma:
@lukefromdc as I said we need to apply this https://github.com/mate-desktop/mate-desktop.org/commit/6a6aa9f21e20760b4260f6795f733753ddede120 to every repo to fix this issue

capplets/appearance/appearance-style.c
lukefromdc: Nice improvement here, should allow all of the usable theme customizations (gsettings does not control window border theme in wayfire etc) to be applied using less code

lukefromdc:
https://github.com/mate-desktop/mate-control-center/pull/758 fixed travis builds in a test

capplets/appearance/appearance-style.c
lukefromdc: https://github.com/mate-desktop/mate-control-center/pull/757/commits/b29a2e863965cea9f161221835b652758c95f80b switches the logic here to exclude the window border theme keys (which are not used in wayfire or similar compositors) and apply the others

capplets/appearance/appearance-style.c
```diff
@@ -152,6 +153,31 @@ treeview_selection_changed_callback (GtkTreeSelection *selection, guint data)
...
+          {
+            if (schema)
+            {
+              GSettings *interface_gnome_settings = NULL;
+              interface_gnome_settings = g_settings_new_full (schema, NULL, NULL);
+
+              /*Exclude marco theming under wayland, the compositor handles window border theming*/
+              if ((strcmp (key, MARCO_THEME_KEY) != 0) &&
+                 (strcmp (key,  WINDOW_TITLE_FONT_KEY) != 0) &&
+                 (strcmp (key,  WINDOW_TITLE_USES_SYSTEM_KEY) != 0))
```
cwendling, line 174: I'm not sure I'm a fan of excluding items instead of including them, but if you think it's better OK

capplets/appearance/appearance-style.c
lukefromdc: Oh yeah-that was just test code, thanks for catching it.

capplets/appearance/appearance-style.c
lukefromdc: Intention here was basically to apply all the possible options save those we know are not used under wayland. Alternately we could include gtk theme, icon theme, cursor theme, cursor size if you feel this is a better approach

lukefromdc: https://github.com/mate-desktop/mate-control-center/compare/b29a2e863965cea9f161221835b652758c95f80b..fc7ec90d9a28031334ccb4c11f16a086e1ed232a

restores the logic of checking for desired conditions rather than excluding.

lukefromdc: Note that while we do have a cursor size key in GNOME, it does not seem to apply from here. I am about to head out on a major road trip so time will be limited

lukefromdc: Found the cursor-size issue: this is not applied from `treeview_selection_changed_callback()` at all, since this key is an integer and not a string. I will need to find out where that is changed, and apply it from there

cwendling: > restores the logic of checking for desired conditions rather than excluding.

I like that :+1:

mark-herbert42:
tested cursor scheme/size - works like a charm.

lukefromdc:
I found the place where cursor size changes are applied and was able to change the GNOME gsettings value from there.

Unfortunately, I also found that Wayfire does not use gsettings to set cursor theme or size at all unless this changed in some recent commit. This the the real reason I am not seeing any response to changing cursor theme or size via gsettngs.I can change it just fine from WCM (wayfire configuration manager) but it appears some other compositors do support changing the cursor size and theme from gsettings.

 At least some versions of sway have been reported to be able to read cursor size and theme from gsettings
https://www.reddit.com/r/swaywm/comments/qq74gv/how_do_i_set_correctly_the_cursor_theme/?rdt=47524
Presumably so does GNOME, meaning some wayland compositors do support cursor theming and size via the GNOME gsettings..

My vote is for adding a warning dialog to indicate that some compositors may ignore gsettings for cursor theme and size, and include cursor theme and size support in this.

thoughts?

lukefromdc:
Also see 
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/58
concerning cursor size and theme setting.
Again, so long as any compositor reads these from gsettings we need to support that, but we also need to warn users on compositors that do not

mark-herbert42:
I use wayfire - git snapshot actually.  But I did not set up any exotic plugins except of rounding corners over it. And - it works for me.  How - I do not know, maybe it is done by gtk through some wayland-wlroots call, but it works. 

And my wayfire is controlled by ini file - not by gsettings, 

Ok - writing this mail I got the point how it works.  So - when mu cursor is hoverinhg over gtk window (or over Qt window ) - it pains cursor theme from org.gnome gsettings. When I hover over window decorations  or Xorg(Xwayland) window it takes wayfire ini file settings. So it actually works excatly as Mate+Compiz in Xorg session.   Gtk overrides the compositor settings. Or WM settings.  So f you set up some fancy cursor in Wayfire.ini but have some legacy in gsettings it will behave inconsistently.   And it is not a wayland-introduced bug (as I had same effect with compiz) - it is just how Mate works with external wm or compositor.

And I do not see many solutions here. Well - solution 1 - is to make a Wayfire add-on like we had on compiz - "Mate-compatibilty" that will read gsettings cursor theme and force it over the one from ini file. or make add-on code for mate-settings - if wayfire detected (display = wayland + wayfire.ini is writeable ) - duplicate settings to wayfire.ini.   Wayfire addon seems to be better way as it is going same ditrection as compiz was doing.

Anyway cursor settings in org.gnome are needed because no matter how wayfire is acting now the org.gnome settings are respected. Over web-browsers, libreoffice or caja-painted desktop org.gnome cursor will be.

lukefromdc:
Just confirmed that even though wayfire has its own ability to specify the cursor, the GNOME gsettings is in fact applied over a GTK window. Changing size works on my setup, the old hardcode I put somewhere for a decade ago to keep early QT4 apps from changing the cursor still means the theme change is visible only in dconf-editor. Anyway, we now have two tests confirming this part of the code works in and only in GTK apps.

lukefromdc:
Does anyone else want to review this before merging? We have a successful test by the reporter of the original issue, and all the travis checks pass

PR #758, opened 2024-06-05
Fix travis deployment

lukefromdc:
Fix travis builds on Fedora and deployment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,8 @@ notifications:
 
 deploy:
   - provider: pages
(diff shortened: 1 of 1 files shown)


lukefromdc:
https://travis-ci.community/t/builds-suddenly-fail-with-error-installing-dpl/14270/4

mbkma (approved):
I would suggest that you can add small changes like this directly to the main branch, due to lack of reviewers. If something goes wrong, we can revert the commit anyway.

PR #766, opened 2024-10-27
keyboard: option to show/select extra layouts

oprudkyi:
- add new checkbox "Include less-common layouts in the selections list" 
image

- switch /org/mate/desktop/peripherals/keyboard/general/load-extra-items
- it is a port/re-work of https://github.com/linuxmint/cinnamon-control-center/commit/d961517fd248f62dc1dc8040731c48822dd68891#diff-7525274c9dc606950625f46b7d8150e4223d7ad4eb9fae59d700d45539c4ccddR196

purpose of the feature. 
it is really hard to know what option controls that extra layout list, I had spent hours groking why my laptop can show them and desktop can't 
so probably it will be useful for newbies and such 

P.S. please be extra careful with review - this is my first ever PR into gtk/mate code

### DIFF
--- a/capplets/keyboard/mate-keyboard-properties-dialog.ui
+++ b/capplets/keyboard/mate-keyboard-properties-dialog.ui
@@ -830,6 +830,22 @@
                             <property name="position">1</property>
                           </packing>
                         </child>
+						<child>
+						  <object class="GtkCheckButton" id="chk_load_extra_items">
+							<property name="label" translatable="yes">_Include less-common layouts in the selections list</property>
+							<property name="visible">True</property>
+							<property name="can_focus">True</property>
+							<property name="receives_default">False</property>
+                            <property name="halign">start</property>
+                            <property name="use-underline">True</property>
+							<property name="draw_indicator">True</property>
+						  </object>
+						  <packing>
+							<property name="expand">False</property>
+							<property name="fill">False</property>
+							<property name="position">2</property>
+						  </packing>
+						</child>
                         <child>
                           <object class="GtkCheckButton" id="chk_separate_group_per_window">
                             <property name="label" translatable="yes">_Separate layout for each window</property>
@@ -843,7 +859,7 @@
                           <packing>
                             <property name="expand">False</property>
                             <property name="fill">False</property>
-                            <property name="position">2</property>
+                            <property name="position">3</property>
                           </packing>
                         </child>
                         <child>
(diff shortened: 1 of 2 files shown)


oprudkyi:
Hi @lukefromdc @oz123 @raveit65 

Could you please check this

lukefromdc:
Github seems to be having trouble at the moment. I can verify from a test build that the checkbox works, but
I have no way to tell if there is a difference in how many layouts are loaded as there are so many either 
way. Note that I know only one language and have never used a different keyboard layout myself.

A proper test of this requires knowing one layout that should appear only when the checkbox for less common
layouts is checked. 

Also note that much of the team seems to be taking a break at this time, and it is beyond me to solo maintain all 
of MATE.

lukefromdc (commented):
I can verify from a test build that the checkbox works, but I have no way to tell if there is a difference in how many layouts are loaded as there are so many either way I cannot tell the difference.

Going to the code, I see the checkbox toggles the second argument to xkl_config_registry_load() but I have never worked on this part of the code so I really don't know how it works 

Note that I know only one language and have never used a different keyboard layout myself. Therefore I have no idea which layouts would be less common.

A proper functional test of this requires knowing one layout that should appear only when the checkbox for less common layouts is checked. If that layout appears when and only when the checkbox is toggled, this PR can be presumed to work.
Any bugs would probably be issues in the existing code in libmatekbd we are toggling. 

Also note that much of the team seems to be taking a break at this time, and it is beyond me to solo maintain all of MATE. We really need a review from someone who knows this part of the code.

oprudkyi:
Hi @lukefromdc 
Thank you 

>A proper functional test of this requires knowing one layout that should appear only when the checkbox for less common layouts is checked. If that layout appears when and only when the checkbox is toggled, this PR can be presumed to work.

well, while it doesn't count - it is surely working for me  :)  
though I haven't found where is testing code or if it is exist to add automated tests 

>We really need a review from someone who knows this part of the code.
ok. I understand, it is not urgent fix

vkareh:
@lukefromdc the checkbox triggers reloading the layout list. This code is already in place in `libmatekbd`: https://github.com/mate-desktop/libmatekbd/blob/master/libmatekbd/matekbd-desktop-config.c#L165-L169
```
	config->load_extra_items =
	    g_settings_get_boolean (config->settings,
				   MATEKBD_DESKTOP_CONFIG_KEY_LOAD_EXTRA_ITEMS);
	xkl_debug (150, "load_extra_items: %d\n",
		   config->load_extra_items);
```
and this PR essentially exposes that toggle.

The list of "extra items" is part of libxklavier (an old freedesktop project, still in use in X11-based distros). In Fedora 41 you can see the list in `/usr/share/X11/xkb/rules/base.extras.xml`

PR #769, opened 2025-01-09
Update time-zones.h Kiev->Kyiv

sggua:
https://en.wikipedia.org/wiki/KyivNotKiev


lukefromdc (approved):
This works without any issues, I had to ensure the string change didn't break whatver computes the time in different zones and it did not. 

Again a one word change to a string, very simple

PR #770, opened 2025-01-09
Update mate-control-center.pot Kiev->Kyiv

sggua:
https://en.wikipedia.org/wiki/KyivNotKiev

### DIFF
--- a/mate-control-center.pot
+++ b/mate-control-center.pot
@@ -4334,7 +4334,7 @@ msgid "Europe/Kaliningrad"
 msgstr ""
 
 #: capplets/time-admin/src/time-zones.h:364
-msgid "Europe/Kiev"
+msgid "Europe/Kyiv"
 msgstr ""
 
 #: capplets/time-admin/src/time-zones.h:365


lukefromdc (approved):
Entirely in favor of this. No way it can break anything, and place names given by foreign occupiers should be replaced with the names given by their rightful occupants

lukefromdc:
I just cherrypicked both of these to 1.28, note that if someone doesn't fix or remove the problem in one translation that's been breaking builds in most repos for most of the year if new translations are pulled from Transifix (which I do NOT know now to do), this one won't be made use of in new releases because it won't be possible for me to update translations.

raveit65:
> I just cherrypicked both of these to 1.28, note that if someone doesn't fix or remove the problem in one translation that's been breaking builds in most repos for most of the year if new translations are pulled from Transifix (which I do NOT know now to do), this one won't be made use of in new releases because it won't be possible for me to update translations.

First you need to generate a new .pot file with new translation strings with `makepot` command.
Commit that and push it to github.
Transiflex check for updates to this file https://github.com/mate-desktop/mate-control-center/blob/master/help/mate-control-center.pot several times a day and load changes automatic.
Than you need to wait some days or weeks for translators do translate new strings.
Updated translations you can pull in with the transifex-client.
`tx pull -af` a=all languages and f=force
I recommend to test building of the package locally before you push changes to github.
Sometimes translators do make format mistakes which can break the build.

lukefromdc:
Translator mistakes ARE the problem that broke updating translations in releases. I have 
no idea how to work on that. Never was that my responsability in the past
but we'be bled off too many team members.

I know how to pull the updates and with a lot of help was able to get translations to work
for mate-wayland-session. I have no idea how to deal with problems on the far end 
though.

raveit65:
I did simply revert an updated language if the language break building.
For example, when the diff pulls in 4 languages and one did break the build, i reverted the commit and pulled in the other 3 Languages with the -L option for language.
And some times you see the formmating issue in po file without understanding the language.
Spending a lot of times alone on such issues was one reason why i leaved mate.

PR #774, opened 2025-03-25
Fix libsystemd auto-detection

cwendling:
Properly handle missing libsystemd when enable-systemd=auto, which is even the default for this switch.

Without this, one had to explicitly give `--disable-systemd` despite the what the help text stated.

---

The reason is that `PKG_CHECK_MODULES()` errors-out if it is not given a 4th argument (action-is-not-found).  The ideal solution for this is however to call with the right arguments depending on whether we want a hard or soft failure, because in case of hard failure `PKG_CHECK_MODULES()` gives a more useful error message than what we would do ourselves.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -179,16 +179,11 @@ AS_IF([test "x$enable_appindicator" = xyes],
 
 have_systemd=no
 AC_ARG_ENABLE(systemd, AS_HELP_STRING([--disable-systemd], [disable systemd support]),enable_systemd="$enableval",enable_systemd=auto)
-if test "x$enable_systemd" != "xno"; then
-	PKG_CHECK_MODULES(SYSTEMD, [libsystemd], [have_systemd=yes],
-	                  [PKG_CHECK_MODULES(SYSTEMD, [libsystemd >= $SYSTEMD_REQUIRED],
-	                  [have_systemd=yes])])
-	if test "x$have_systemd" = xno && test "x$enable_systemd" = xyes; then
-		AC_MSG_ERROR([*** systemd support requested but libraries not found])
-	elif test "x$have_systemd" = xyes; then
-		AC_DEFINE(HAVE_SYSTEMD, 1, [Define if systemd is available])
-	fi
-fi
+AS_CASE(["$enable_systemd"],
+        [yes], [PKG_CHECK_MODULES(SYSTEMD, [libsystemd >= $SYSTEMD_REQUIRED], [have_systemd=yes])],
+        [auto], [PKG_CHECK_MODULES(SYSTEMD, [libsystemd >= $SYSTEMD_REQUIRED], [have_systemd=yes], [have_systemd=no])],
+        [have_systemd=no])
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
An autotools test build on a systemd setup successfuly detected the presence of systemd, I don't have any non-systemd systems to test the absence of it on though, 

Works for the systemd present case

lukefromdc:
@cwendling , I am fine with merging this if you had a succesful autotools build (we also have meson support) on a machine without systemd. Also OK with waiting for another review if you prefer. Please advise

cwendling:
@lukefromdc thanks for the test.  Though, if you want to test the non-systemd state, you simply need to remove the systemd *development files*, not systemd itself.  IIRC on Debian & Ubuntu it's the `libsystemd-dev` package, no idea on Fedora but maybe `systemd-devel` or so -- but you'd probably know.

lukefromdc:
removing libsystemd-dev did NOT work, I got

```
Package 'libsystemd', required by 'dbus-1', not found
Package 'libsystemd', required by 'dbus-1', not found
Package 'libsystemd', required by 'dbus-1', not found
Package 'libsystemd', required by 'dbus-1', not found

```
Cannot remove anything else for this test, remember this is a live system not a VM

lukefromdc (commented):
Turning my approval to a comment due to the failed build with libsystemd-dev removed

cwendling:
@lukefromdc What is the distro you used?  On my Debian Bookworm it clearly works: with `libsystemd-dev`, it works OK without this patch, but if I remote it, master gives me:

```console
$ ./configure
[…]
checking for libsystemd... no
checking for libsystemd >= 248... no
configure: error: Package requirements (libsystemd >= 248) were not met:

Package 'libsystemd', required by 'virtual:world', not found
[…]
```

while with this patch I get:
```console
$ ./configure
[…]
checking for libsystemd >= 248... no
[…]
Configure summary:

        mate-control-center 1.29.0
        ==========================
[…]
	Systemd:                           no
[…]
```

lukefromdc:
I am on a rather modified version of Debian Unstable

cwendling:
Hum… can you tell what depends on dbus-1?  I can't seem to find a reference to it in the build system, and if I check the pkg-config result for dbus-1 I don't see any reference to libsystemd -- and if there is one in a newer version, it should have the associated dependency, if not that's a packaging issue.

This said, what I tried to fix was the *explicit* and optional dependency on libsystemd we have, that could be disabled before my patch by explicitly asking for it.  If there's indirect dependencies on it (e.g. through the dbus-1 pkg-config file), it's outside the scope of this PR, and I'm not interested in fighting those :)

lukefromdc:
I don't really have an easy way to figure that out. I am basically not in position to provide effective support for the non-systemd case and I don't want to make changes to a running system that would prevent restarting if I get a crash

vkareh (approved):
Yeah, I get `Package 'libsystemd' not found` on Fedora 42, and the whole thing dies. This fixes it, thanks!

PR #775, opened 2025-03-25
Improve accessibility of the keybindings properties dialog

cwendling:
* Hint to the keyboard navigation in the help text
* Present something to the screen reader when waiting for a new shortcut
* Fix the value screen readers see for the shortcut of group rows (which is visually empty, but had seemingly random values for the screen readers)

@joanmarie @shindere Hope that fixes https://www.freelists.org/post/orca/Could-Orca-be-more-accurate-in-the-keyboard-shortcut-system-window,4  Could you give it a spin?

However, I couldn't seem to get the help-text stuff to work, even with Orca 48/AT-SPI 2.56 from Debian Bookworm's backports.
I hope the updated help text and tooltip are enough; and there's an additional fix for group rows.

### DIFF
--- a/capplets/keybindings/eggcellrendererkeys.c
+++ b/capplets/keybindings/eggcellrendererkeys.c
@@ -11,6 +11,8 @@
 
 #define TOOLTIP_TEXT _("New shortcut...")
 
+static void update_text (EggCellRendererKeys *keys);
+
 static void             egg_cell_renderer_keys_finalize      (GObject             *object);
 static void             egg_cell_renderer_keys_init          (EggCellRendererKeys *cell_keys);
 static void             egg_cell_renderer_keys_class_init    (EggCellRendererKeysClass *cell_keys_class);
@@ -76,6 +78,8 @@ GType egg_cell_renderer_keys_get_type(void)
 
 static void egg_cell_renderer_keys_init(EggCellRendererKeys* cell_keys)
 {
+	g_signal_connect (cell_keys, "notify::visible", G_CALLBACK (update_text), NULL);
+
 	cell_keys->accel_mode = EGG_CELL_RENDERER_KEYS_MODE_GTK;
 }
 
@@ -223,6 +227,27 @@ static gchar* convert_keysym_state_to_string(guint keysym, guint keycode, EggVir
 	{
 		return egg_virtual_accelerator_label(keysym, keycode, mask);
 	}
+}
+
+static void update_text (EggCellRendererKeys *keys)
+{
+  gboolean visible;
+
+  g_object_get (keys, "visible", &visible, NULL);
+
+  /* For screen reader's sake, make the text empty when hidden, otherwise a
+   * more or less random value is presented.  This could also be fixed in the
+   * accessibility implementation side, but we rely on GTK's, so it's easier
+   * to clean it up here than modify GTK or subclass it further. */
+  if (! visible)
+    g_object_set (keys, "text", "", NULL);
+  else
+  {
+    /* sync string to the key values */
+    gchar *text = convert_keysym_state_to_string (keys->accel_key, keys->keycode, keys->accel_mask);
+    g_object_set (keys, "text", text, NULL);
+    g_free (text);
+  }
 }
 
 static void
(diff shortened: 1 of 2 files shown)


lukefromdc (commented):
The help text at the bottom of the capplet works fine-and I finally got Orca to work! This does indeed fix 
https://www.freelists.org/post/orca/Could-Orca-be-more-accurate-in-the-keyboard-shortcut-system-window,4
the updated text is read immediately

One issue is this: when the active row is a group like "Desktop" we get things like "tree level 1" read rather than no sound, which is what we would expect with an empty label

shindere:
> The help text at the bottom of the capplet works fine-and I finally got Orca to work! This does indeed fix 
> https://www.freelists.org/post/orca/Could-Orca-be-more-accurate-in-the-keyboard-shortcut-system-window,4

Thanks for your work!

> One issue is this: when the active row is a group like "Desktop" we
> get things like "tree level 1" read rather than no sound, which is
> what we would expect with an empty label

Is it an issue really?

I mean, is it indeed wrong htat a group is a tree that can be either
closed or opened? If indeed it is not the case that this is a tree, then
indeed announcing is as such feels misleading to me.

Also, would it make sense to change the UI so that thereisfirst a list
from which one picks the category, and then in the following field the
table tiwh just the sohrtcuts of this category?

On my system the whole table (with the category titles) has more than
hundreed elements, which feels huge and I wouln't mind having this
broken into several smaller pieces that are not displayed
simultaneously.

lukefromdc:
I really don't know if that's an issue or not, not being an end user of Orca

lukefromdc:
@cwendling , are you OK with merging this as it stands now?

cwendling:
>> I don't thing we're not doing anything wrong, and I don't think there's any reason to change the layout here.
>
> A lot of negations here, I think you meant "we're doing anything wrong", right?

Oops, there was one "not" too many.  I edited the comment, but I meant: "I think that what we are doing is fine and common, no need to change it unless there are specific and motivated reasons to do so."

>> One relatively simple improvement if the list size is getting too large could be a "smart" search filter that would reduce the amount of visible items based on the search terms. Yet again, I'm not sure it's worth the trouble as it stands, but other users can give their opinion on this.
>
> Not found of things who try to be intelligent, TBH.

By "intelligent" I simply meant "not too dumb", e.g. perform substring marching or other form fuzzy matching, or even introduce keywords that can match, so users can actually find what they are looking for without having to know exactly how it's worded.  No LLM in sight, don't worry :grin: 

> I did try and, as I reported on the Orca mailing list, I did notmanage to build.

Oh, I missed that.  Will try to find the mail and give you a hand.

> If nobody else stands up I think why not merge it as it is so that it gets downstream and thus becomes easier to test by more users and we can iterate from there if needed, likely with more feedback.

Sure, and that will happen eventually if we don't get more testers, but it's always nice to test changes ahead not to create problems for more innocent users :slightly_smiling_face:

lukefromdc:
Merged as it works here and the tester we were waiting on wasn't able to build mate-control-center. This is one of those things that has to work or people get excluded at the end user level

cwendling:
@shindere cool, thanks for the confirmation!

> I believe that this dialog may still benefit from help texts that are more easily consumable by screen-reader users but this can certainly come later.

Suggestions are welcome :)  I'd likely be fine implementing specific things you think should be adjusted -- so long as I find time for it :slightly_smiling_face:

shindere:
Colomban Wendling (2025/04/08 00:35 -0700):
> @shindere cool, thanks for the confirmation!
My pleasure

> Suggestions are welcome :)

The thing is, I do not know where there are possibilities to add /modify
things.

What I know is that the text that is aanounced when the dialog box opens
assumes you have a mouse as it said to reach the line of interest and
"double-click" on it. Tome this is an accessibility issue that is beyond
just visually impaired persons.

> I'd likely be fine implementing specific things you think should be
> adjusted -- so long as I find time for it :slightly_smiling_face:

Thanks a lot. As I said abouve I fear I do not know which possibilities
are available so it's hard for me to suggest something, but we can
exchange offline andtalk to each other, maybe both of us will manage to
come up with something.

shindere:
Colomban Wendling (2025/04/08 07:31 -0700):
> I modified this message to mention keyboard as well: *double-click **or press Space or Enter*** (emphasis mine).
> Is that not enough, or doesn't this get through to you?  It did to me.

I may have forgotten that change, sorry.

I think once you are on the right line you need to arrow right before
pressing Enter?

cwendling:
> I think once you are on the right line you need to arrow right before pressing Enter?

Not necessarily no, it works on any columns from what I saw.

shindere:
Colomban Wendling (2025/04/08 07:44 -0700):
> cwendling left a comment (mate-desktop/mate-control-center#775)
> 
> > I think once you are on the right line you need to arrow right before pressing Enter?
> 
> Not necessarily no, it works on any columns from what I saw.

Okay then it sounds fine in its current state, many thanks for your help
on this.

Issue #776, opened 2025-04-29
Duplicate and disappearing buttons in MATE Font Viewer


MrMinderbinder:
The back button duplicates itself each time you view a font in Font Viewer without closing it. It disappears completely the fourth time. This is with MATE 1.26.0.

![Image](https://github.com/user-attachments/assets/aec4c28b-71f3-44e1-b6b3-8760f396c1f3)
![Image](https://github.com/user-attachments/assets/3dea8131-fb2e-4272-bd09-c0aa148e26a4)

lukefromdc:
How are you doing that? The only way I was able to get back to the list of fonts to view another was with the back button, and I did not get this problem in so doing

MrMinderbinder:
I opened the fonts by right clicking on the font file in Caja and selecting “Open With Mate Font Viewer” from the context menu. Doing this repeatedly with either the same file or a different one produces this result.

lukefromdc:
OK, this is CONFIRMED. My guess is when it was gnome-font-viewer the original authors didn't consider this use case,
or maybe something in the GTK2->GTK3 transition did this.

carsonhicks:
I looked into this bug and found that it was also causing duplication of the Info/Install buttons. The disappearing buttons appears to be related to the toolbar running out of space, as resizing the window on a large monitor allowed me to see more back buttons.

cwendling:
Fixed in a54d97435fd3f53d803aa5230148ddc024e312df

PR #777, opened 2025-05-14
Fix duplicate buttons in font viewer

carsonhicks:
This fixes the behavior reported in #776 by checking to make sure the toolbar buttons for Install/Info/Back don't already exist before creating them.


cwendling:
Merged as a54d97435fd3f53d803aa5230148ddc024e312df, with a couple whitespace fixes.  Thanks!

PR #778, opened 2025-05-14
maybe fix Travis

cwendling:


cwendling:
@raveit65 thanks for the hint, I'll give this a look

yetist:
Maybe we can try to use the github action?

mbkma:
we now have gh actions

PR #781, opened 2025-05-16
Use github actions build

yetist:
Try to use github action build


oz123:
Looks good to me. A good start for removing Travis.

oz123 (approved):

oz123 (changes requested):
Actually, @yetist can we change the build to use meson instead of auto tools?

capplets/display/meson.build
```diff
@@ -10,7 +10,13 @@ sources += gnome.compile_resources(
   export: true
 )
 
-sources += gnome.genmarshal('foo-marshal', prefix : 'foo_marshal', sources : 'foo-marshal.list')
+marshalers = gnome.genmarshal(
+'foo-marshal',
```
oz123, line 14: What is foo-marshal for?

oz123 (commented):
Great work. Still need to review everything!

capplets/display/meson.build
yetist: No changes are needed here. Previously, meson compilation failed because the generated files were compiled with autotools first. Now compile meson first, then compile autotools.

yetist:
I think it's ready for review now.
The release test is here: https://github.com/yetist/mate-control-center/releases

zhuyaliang (approved):
good

lukefromdc (approved):
A meson test build of the release tarball built fine and ran fine locally on Debian Unstable

Issue #785, opened 2025-05-29
Invalid layout


luigifab:
Compared to `mate-appearance-properties`, `mate-default-applications-properties`, `mate-keyboard-properties`, `mate-network-properties`, `mate-mouse-properties`, `caja-file-management-properties`, *print dialog*, there is a "bug" with position of the main notebook with:

**mate-window-properties**

![Image](https://github.com/user-attachments/assets/4474efde-6786-4480-8697-dacd2d90c379)

https://github.com/mate-desktop/mate-control-center/blob/4c50efe0ee15c9dfa45fc26c8b75ecbef4da12bc/capplets/windows/window-properties.ui#L47-L65

- GtkDialog : border-width -> 5
- GtkBox : border-width -> 2 / spacing -> 2
- GtkNotebook : border-width -> 5

**mate-power-preferences**

- GtkDialog : border-width -> 5
- GtkNotebook : border-width -> 5 / margin -> 0

**mate-session-properties**

- GtkDialog : border-width -> 5
- GtkNotebook : border-width -> 5

There is also a maximize button, but not for all windows.
To compare layout, open all windows, and put all windows top/left at same place.

Mate 1.26

PR #786, opened 2025-06-02
Update window layout

luigifab:
Fix https://github.com/mate-desktop/mate-control-center/issues/785
Tested by rebuilding Mate 1.26 Debian source package.


lukefromdc:
For me to test this will have to wait, as to get this capplet to open I have to close out my session, log back into the x11 session, then change the WM in it from compiz to marco

PR #787, opened 2025-06-02
Remove focus on close button

luigifab:
This restore focus on first notebook tab title.

Help https://github.com/mate-desktop/mate-control-center/issues/785
Tested by rebuilding Mate 1.26 Debian source package.


lukefromdc (commented):
I still got focus on the close button with this, thinking this may be being set somewhere else in the code as well

luigifab:
loool, right... maybe I was dreaming when I tested it!

lukefromdc:
No idea, while I was NOT testing 1.26 I don't think this would have changed. Yes I restarted the panel
for every test. I have no further ideas about what is different. Not worried about it as I am OK with this
either way

vkareh (approved):
looks good to me

PR #794, opened 2025-09-02
windows: Add UI for titlebar mouse actions

vkareh:
Expose new titlebar mouse actions in the window properties dialog (middle-click, right-click, scroll wheel).

Note: requires https://github.com/mate-desktop/marco/pull/808 to work properly


lukefromdc (commented):
This works fine once the relevent marco commit (now merged to master) is applied.

We probably need to version bump both marco and mate-control-center for this though as it causes mate-control-center to depend on marco with the matching commit applied.

Marco can work fine with the new feature on its own, but would need a minor version bump simply to identify it. We might need a major version bump though for this one, as some distros are already not picking up 1.28 and adding new dependencies in a minor version update will be another problem in this regard.

PR #795, opened 2025-09-07
Update window layout

luigifab:
This old PR align bottom spacing between _notebook_ & _buttons_ for **mate-default-applications-properties** like in **mate-appearance-properties**.

before:

image

---

after:

image


lukefromdc:
In all themes I have EXCEPT the Menta and Submarine themes, these already line up. These two sets of thems pad out the combobox buttons in the notebook and misalign the buttons on the bottom. Will test this now

PR #809, opened 2026-06-09
display: fix preview ghosting when rearranging monitors

### LINKED ISSUE #808: preview screen tearing
### Steps to Reproduce

This has been an issue for a very long time, for me, on every machine without a GPU, when relocating monitors, the icon, in the gui preview doesnt delete the graphic from the previous location. 
I have a picture. It gets confusing, its hard to see where the screen is. closing, and re-opening the app fixes it. 
so something about the drag and drop function doesn't clear the previous positions.

not really life changing either way, but it has done this for as long as I can remember using mate. 

Image

### Expected Behavior

graphic for previous position disappears when new position is selected. 

### Actual Behavior

Both old and new position are shown

### Error Output

```shell
screen shot from mate desktop 1.26.0
```

### Version

1.26.0

### Distribution

linux mate version 1.26.0

### Additional Context

_No response_

### Reproducibility

Every time

### Checklist

- [x] I have searched existing issues to avoid duplicates
- [x] I have provided all the requested information above

vkareh:
Dragging monitors in the display preferences left ghost images of the previous position on systems without a compositor.

This patch removes creating an entire drawing context on the parent window just for testing when edges touch and instead use a single-pixel cairo surface. That way we avoid interfering with pending redraws. Then we replace the manual canvas-to-window coordinate translation for invalidation with a simple redraw call, which should handle windowless widget positioning correctly.

Fixes #808

### DIFF
--- a/capplets/display/scrollarea.c
+++ b/capplets/display/scrollarea.c
@@ -806,7 +806,6 @@ process_event (FooScrollArea	       *scroll_area,
 	       int			x,
 	       int			y)
 {
-    GtkWidget *widget = GTK_WIDGET (scroll_area);
     guint i;
 
     allocation_to_canvas (scroll_area, &x, &y);
@@ -832,12 +831,10 @@ process_event (FooScrollArea	       *scroll_area,
 	    {
 		cairo_t *cr;
 		gboolean inside;
-		GdkDrawingContext *gdc;
-		cairo_region_t    *cairo_region;
+		cairo_surface_t *surface;
 
-		cairo_region = cairo_region_create ();
-		gdc = gdk_window_begin_draw_frame (gtk_widget_get_window (widget), cairo_region);
-		cr = gdk_drawing_context_get_cairo_context (gdc);
+		surface = cairo_image_surface_create (CAIRO_FORMAT_ARGB32, 1, 1);
+		cr = cairo_create (surface);
 		cairo_set_fill_rule (cr, path->fill_rule);
 		cairo_set_line_width (cr, path->line_width);
 		cairo_append_path (cr, path->path);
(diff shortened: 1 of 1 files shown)


cwendling:
I didn't test this, but I don't understand the relation between the 2 parts of the changes:
1. do not use the widget's surface as target for path intersection testing (and use a custom target surface)
2. redraw the whole widget rather than specific regions

Both *could* be the source of drawing issues, but I don't understand how they actually relate.

Also, quickly looking at the code suggests simplifying by redrawing the whole widget rather than a region would likely be fine, but that could actually remove a whole lot more code than just the part in the patch I believe.

vkareh:
So originally I started with the change in `process_event`, since it seemed to me the most obvious place where the tearing could be happening (since it does a full Xserver roundtrip on every pixel the mouse moves). By allocating a 1x1 surface and testing the position directly we skip the roundtrip.

But that didn't fix it until I changed to a full redraw, which is what fixed the tearing for me. I honestly did not test the case of the full redraw without the process event, but also it didn't make sense to remove it since the original code (with the gdk context) would also get in the redraw queue unnecessarily.

vkareh:
> Also, quickly looking at the code suggests simplifying by redrawing the whole widget rather than a region would likely be fine, but that could actually remove a whole lot more code than just the part in the patch I believe.

Possibly, I can see a bunch of offset translations that would probably be redundant if we're doing a full redraw, but not sure what the effect of those would be on the entire scroll area (I honestly don't yet understand why use a scrollable widget here)

lukefromdc (approved):
While I have never seen the issue this is aimed at in many years, it didn't break anything and is actually a simplification of the code as well. I had no issues with it at all and the rebase with the current 1.29/wayland code went fine

lukefromdc:
Cherrypicked to 1.28 as this was no doubt an aggravating bug. Also the whole Travis->github-actions and Transifex->Weblate transitions were not yet in 1.28 so I fixed that too, then tested with `make distcheck ` before pushing the branch, and test ran the build file locally in Xwayland and it worked fine.

PR #813, opened 2026-07-23
default-applications: find apps missing from mimeinfo.cache

### LINKED ISSUE #150: Can't set default application from ~/.local/share/applications
Hello

With mate-default-applications-properties (1.8.3+dfsg1-1):
- I can't set Pluma to default text editor (there are only Geany and LibreOffice Writer)
- I can't set VLC to default video player or multimedia player (the drop down is empty)
- I can't set Firefox (manual installation) or Chomium (debian installation) and Thunderid (manual installation) to default browser or mail reader.

## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9591446-can-t-set-default-application?utm_campaign=plugin&utm_content=tracker%2F559197&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F559197&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vkareh:
Preferred Applications only listed apps that were registered in mimeinfo.cache. Apps installed in unusual ways, or with a stale or missing cache, wouldn't show up as options for web browser, mail reader, and other defaults.

This change now scans each installed .desktop file directly for its MimeType entry, so apps missing from the cache are still found.

Also fixes a longstanding bug where duplicate messenger entries weren't actually being filtered out.

Fixes #150

### DIFF
--- a/capplets/default-applications/mate-da-capplet.c
+++ b/capplets/default-applications/mate-da-capplet.c
@@ -579,6 +579,46 @@ compare_apps (gconstpointer a, gconstpointer b)
 	return ret;
 }
 
+static gint
+app_info_cmp (gconstpointer a, gconstpointer b)
+{
+	return g_app_info_equal (G_APP_INFO (a), G_APP_INFO (b)) ? 0 : 1;
+}
+
+/* Get all apps that support a MIME type by checking both mimeinfo.cache and
+ * the MimeType= field in each .desktop file directly. The cache lookup misses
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
This REALLY works, suddenly a lot more options that would be correct but never appeared before now show up in the default applications comboboxes, librewolf included.

PR #814, opened 2026-08-03
appearance: verify if theme is GTK3

### LINKED ISSUE #276: 3.22.21: gtk murrine engine dependencies and other gtk 2.x dependent themes
Looks like some mate-themes themes depends on gtk murrine engine which is no longer available and only supperts  gtk 2.x.
I thoing that it would be good to make decision about take over gtk murrine engine mainatainance or remove those themes from source tree as mate desktop compeltly swithed to gtk 3.x.

I'm not sure but I thing that currently there are more gtk 2.x parts.

vkareh:
mate-appearance-properties only looked for a gtk-2.0/gtkrc file when deciding whether a theme was installed, so GTK3 themes were reported as not installed.

This adds a monitor for the gtk-3.0/gtk.css file to validate whether the theme is GTK3.

Fixes mate-desktop/mate-themes#276

### DIFF
--- a/capplets/appearance/appearance-style.c
+++ b/capplets/appearance/appearance-style.c
@@ -805,7 +805,7 @@ create_thumbnail (const gchar *name, GdkPixbuf *default_thumb, AppearanceData *d
   } else if (default_thumb == data->gtk_theme_icon) {
     MateThemeInfo *info;
     info = mate_theme_info_find (name);
-    if (info != NULL && info->has_gtk) {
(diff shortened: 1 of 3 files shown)


lukefromdc (approved):
This is simple (regardless of the number of LOC) and it works. I deleted the GTK2 directory from BlackMATE in my somewhat old mate-themes install and this worked fine.

The code change takes every place GTK2 is looked for and also looks for GTK3 and only throws the error message if neither is found.

PR #815, opened 2026-08-06
display: add Wayland capplet backed by the wlrandr plugin

Sunderland93:
Снимок экрана_20260806_212319

Depends on https://github.com/mate-desktop/mate-settings-daemon/pull/439

### DIFF
--- a/capplets/display/Makefile.am
+++ b/capplets/display/Makefile.am
@@ -19,6 +19,7 @@ nodist_mate_display_properties_SOURCES = \
 	$(BUILT_SOURCES)
 mate_display_properties_SOURCES =	\
 	xrandr-capplet.c		\
+	wlrandr-capplet.c		\
 	scrollarea.c			\
 	scrollarea.h
 
--- a/capplets/display/meson.build
+++ b/capplets/display/meson.build
@@ -1,5 +1,6 @@
 sources = files(
   'xrandr-capplet.c',
+  'wlrandr-capplet.c',
   'scrollarea.c',
 )
 
--- a/capplets/display/scrollarea.c
+++ b/capplets/display/scrollarea.c
@@ -479,7 +479,32 @@ foo_scroll_area_draw (GtkWidget *widget,
     FooScrollArea *scroll_area = FOO_SCROLL_AREA (widget);
     cairo_t *cr;
     cairo_region_t *region;
-    GtkAllocation widget_allocation;
+
+    /* The backing store inherits the device scale of the window it was
+     * created from (see create_new_surface).  If the window now lives on
+     * an output with a different scale, recreate it so the blit below
+     * stays 1:1; otherwise only a corner of the preview is drawn.
+     */
+    {
+	cairo_surface_t *target = cairo_get_target (widget_cr);
+	double tx, ty, sx, sy;
+
+	cairo_surface_get_device_scale (target, &tx, &ty);
+	cairo_surface_get_device_scale (scroll_area->priv->surface, &sx, &sy);
+
+	if (tx != sx || ty != sy)
+	{
+	    cairo_surface_t *new;
+	    GtkAllocation allocation;
+
+	    gtk_widget_get_allocation (widget, &allocation);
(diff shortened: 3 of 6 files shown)


lukefromdc:
In testing this was unable to detect my monitor, though the UI came up most of it was greyed out. Got this error:
error

Also could not show a tray icon, but that will have to be ported to SNI to render under wayland

Sunderland93:
icon is already wayland compatible and can be enabled via `gsettings set org.mate.SettingsDaemon.pugins.wlrandr show-notification-icon true` if there is problem with UI. Seems like wlrandr schema is incorrectly install

lukefromdc:
I just found /usr/share/glib-2.0/schemas/org.mate.SettingsDaemon.plugins.wlrandr.gschema.xml in its intended place, and did recompile the schemas

The error message has the capplet looking for `/org/mate/SettingsDaemon/WLRANDR`

while dconf-editor shows `/org/mate/settings-daemon/plugins/wlrandr` existing, same as the other plugins

lukefromdc:
Running `gsettings set org.mate.SettingsDaemon.plugins.wlrandr show-notification-icon true` completes without an error message, and dconf-editor shows a reload for the page shown below  but the SNI icon does not appear
dconf-editor

Sunderland93:
I see no errors on my side. Will check on fresh setup. Also please ensure that wlrandr plugin is running and registred:

`gdbus introspect --session --dest org.mate.SettingsDaemon --object-path /org/mate/SettingsDaemon/WLRANDR`

Also make sure you're running mate-settings-daemon as Wayland native, without GDK_BACKEND=x11

lukefromdc:
looks like mate-wayland-session needs a change to remove  `GDK_BACKEND=x11` from
`restart_while_running env GDK_BACKEND=x11 mate-settings-daemon &`
That commit will have to depend on this PR to work

lukefromdc:
That was IT! The capplet and the icon now work fine. Since the wayland session auto-restarts mate-settings-daemon I was unable to get the restart to run under native wayland until I removed that line. Good to go

lukefromdc:
Only remaining issue I am finding with this PR is the image of the screen not centered and the right edge cut off with a 4K monitor

Display_Capplet

lukefromdc:
We have the two issues I just noted with the clipboard and xsettings plugins over at 
https://github.com/mate-desktop/mate-settings-daemon/pull/439
Once those are fixed and that is merged, this is probably OK to merge as-is and deal later with the drawing of the screen rectangle (not critical to use of this capplet) or that can be fixed here and now if that fix is easily found. That has to merge first as the capplet depends on it to function

Sunderland93:
Latest commit should resolve display image issue

lukefromdc:
This is what I just got after the latest commit, looks like the image is just getting cut off on the right side edge

wlrandr

Sunderland93:
Hm, I can't reproduce it on Labwc

Снимок экрана_20260814_221056

Sunderland93:
> This is what I just got after the latest commit, looks like the image is just getting cut off on the right side edge
> wlrandr

I just noticed that in your screenshot, the display settings window only has a close button. Does this window not resize on your system? This could explain why the display preview widget doesn't fit into the existing layout.

Sunderland93:
Hm. I think it works on Labwc because it ignore window constraints. Will check on Wayfire

Снимок экрана_20260814_224402

lukefromdc:
The display properties window can be resized larger but not smaller, and internal contents do not resize on
my setup

Sunderland93:
Okay, I confirm that this problem occurs on Wayfire

Sunderland93:
Latest commit fixes issue in my Wayfire setup

lukefromdc (approved):
This now works perfectly, though it depends on https://github.com/mate-desktop/mate-settings-daemon/pull/439 so that has to be merged first.

lukefromdc:
Translation PR passed checks. will release 1.29.1 with this and updated translations

